PR #5, opened 2016-04-19
Topic pages - add an accordion with descriptions

gemmaleigh:
Add an accordion with descriptions module to enhance subsections - a heading, description and list of links - when JavaScript is enabled.

This module has been [added to the GOV.UK accessibility sandbox](http://govuk-accessibility-sandbox.herokuapp.com/patterns/accordion-with-descriptions/02/).

**Accordion with descriptions**

_What is it?_

The pattern is a list of headings each with a description and some hidden content. When you select the heading or the description the hidden content is revealed. There is also a button to open and close all the content.

The pattern allows users to see a summary of available content which can then be expanded if a user wants to read more.

**With JavaScript disabled:**

js-disabled

**With JavaScript enabled:**

agile delivery gov uk

agile delivery gov uk - open


config/initializers/assets.rb
```diff
@@ -10,3 +10,10 @@
   application-ie7.css
   application-ie8.css
 )
+
+# Precompile js
+Rails.application.config.assets.precompile += %w(
+  govuk/modules.js
+  modules/accordion-with-descriptions.js
+  start-modules.js
+)
```
thehenster: I don't think we need to precompile these if we're including the whole tree in application.js with `//= require_tree .` and they're in app/assets.

spec/javascripts/accordion-with-descriptions-spec.js
```diff
@@ -0,0 +1,181 @@
...
+  it("has a an aria-expanded attribute and the value is false", function () {
+    var $subsectionButton = $element.find('.subsection__title button:first');
+    var $subsectionContent = $element.find('.subsection__content');
+    $subsectionButton.click();
+    expect($subsectionButton).toHaveAttr('aria-expanded','true');
+    $subsectionButton.click();
+    expect($subsectionButton).toHaveAttr('aria-expanded','false');
+  });
+
+});
```
thehenster: 👏  for the tests

spec/javascripts/accordion-with-descriptions-spec.js
```diff
@@ -0,0 +1,181 @@
...
+
+  // Set the text inside the button to "Open all"
+  it("has an open/close all button with text inside which is equal to Open all", function () {
+    var $openCloseAllButton = $element.find('.js-subsection-controls button');
+
+    expect($openCloseAllButton).toHaveText("Open all");
+  });
+
+  // Ensure that the value for the aria-controls attribute matches the IDs of all of the subsections
+  it("has a value for the aria-controls attribute which includes all of the subsection_content_IDs", function () {
```
thehenster: I find the comment doesn't add much value on top of the example description here. I'd just bin the comment?

spec/javascripts/accordion-with-descriptions-spec.js
```diff
@@ -0,0 +1,181 @@
...
+  });
+
+  // Ensure that the subsection-icon div has been inserted
+  it("has a header with a child element which has a class of .subsection-icon", function () {
+    var $subsectionHeader = $element.find('.subsection__header');
+
+    expect($subsectionHeader).toContain('.subsection__icon');
+  });
+
+  // When all sections are closed, make sure there are no .is-open classes
```
thehenster: I like all the explanation of what is going on. I think this type of comment can work well as another `describe` block so it becomes a little more "alive" and more likely to stay updated. Also, when it's a describe/context block it annotates the test runner output which tells a more complete story which is nice (sometimes).

spec/javascripts/accordion-with-descriptions-spec.js
```diff
@@ -0,0 +1,181 @@
...
+            <div class="subsection__content" id="subsection_content_1">\
+              <ul class="subsection__list">\
+                <li class="subsection__list-item">\
+                  <a href="">Subsection list item in here</a>\
+                </li>\
+              </ul>\
+            </div>\
+          </div>\
+        </div>\
+      </div>');
```
thehenster: Only a minor thing.. Do we need all the backslashes? It makes it just a little tiny bit harder to read and wouldn't having the markup over multiple lines to closer to real life anyway?

thehenster:
I've made a couple of smallish comments. Looks good :+1:

spec/javascripts/accordion-with-descriptions-spec.js
gemmaleigh: We do need those, to escape the line breaks, so leaving in.

config/initializers/assets.rb
gemmaleigh: I've removed this extra config.

gemmaleigh:
🎆

PR #8, opened 2016-04-28
Display a lede with the guide summary

thehenster:
This is being implemented for the point page that uses the service_manual_guide schema.

(The build will be broken until https://github.com/alphagov/govuk-content-schemas/pull/294 makes it's way in)


AndrewVos:
Looks good, merge it when the tests pass.

PR #10, opened 2016-05-04
Optionally collapse topics

thehenster:
We are now allowing the content designers in the publisher to decide whether the accordion is applied or not. There wasn't a clear set of rules that we could define. It seems as though a human needs to decide whether it is collapsed or not.

The default is that the topic sections are expanded. It is more likely that a topic will be collapsed but the downside of a difficult to use topic is more significant than the upside of saving a click in the publisher.

### DIFF
--- a/app/views/content_items/service_manual_topic.html.erb
+++ b/app/views/content_items/service_manual_topic.html.erb
@@ -22,7 +22,11 @@
       <%= @content_item.description %>
     </p>
 
-    <div class="js-hidden" data-module="accordion-with-descriptions">
+    <% if @content_item.visually_collapsed? %>
+      <div class="js-hidden" data-module="accordion-with-descriptions">
+    <% else %>
+      <div>
+    <% end %>
       <div class="subsection-wrapper">
 
         <% @content_item.groups.each_with_index do |link_group, idx| %>
--- a/test/integration/service_manual_topic_test.rb
+++ b/test/integration/service_manual_topic_test.rb
@@ -39,4 +39,17 @@ class ServiceManualTopicTest < ActionDispatch::IntegrationTest
         href: "/service-manual/communities/user-research-community")
     end
   end
+
+  test "it does not apply the accordion if the topic isn't visually collapsed" do
+    setup_and_visit_content_item('service_manual_topic')
+
+    refute page.has_css?(%{div[data-module="accordion-with-descriptions"]})
+  end
+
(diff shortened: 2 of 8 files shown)


app/views/content_items/service_manual_topic.html.erb
```diff
@@ -22,7 +22,11 @@
       <%= @content_item.description %>
     </p>
 
-    <div class="js-hidden" data-module="accordion-with-descriptions">
+    <% if @content_item.visually_collapsed? %>
```
AndrewVos: Realllllly don't like this approach here where we're outputting only one part of the div. I usually like to do something like:

```erb
<% container_class = @content_item.visually_collapsed? ? "js-hidden" : "" %>
<div class="<%=container_class%>">
</div>
```

```

Comment (AndrewVos) [inline app/views/content_items/service_manual_topic.html.erb]: I'm not sure why I don't like it though. Probably cause it makes nesting weird?

File: test/integration/service_manual_topic_test.rb [hunk 2]
```diff
@@ -39,4 +39,17 @@ class ServiceManualTopicTest < ActionDispatch::IntegrationTest
         href: "/service-manual/communities/user-research-community")
     end
   end
+
+  test "it does not apply the accordion if the topic if it isn't visually collapsed" do
```
AndrewVos: this test name is odd. Use your words :)

app/views/content_items/service_manual_topic.html.erb
AndrewVos: I just noticed the `accordion-with-descriptions` part hah.

AndrewVos: Why are you only appling `accordion-with-descriptions` if it's collapsed? Doesn't seem obvious to me.

thehenster: Yeah.. I nearly went for a `<%= content_tag :div, some_hash_I've_sculpted %>` but it still needed me to define a hash with a condition that looked like.. `{class: "js-hidden", data: { module: "accordion-with-descriptions" } }`. It was less readable :( I didn't really want to push it into a helper either as it would hide what it's up to.. sigh.

The accordion is the thing that does the collapsing.. If it's expanded then we don't want the accordion.

test/integration/service_manual_topic_test.rb
thehenster: Ha - thanks

PR #11, opened 2016-05-04
[DO NOT MERGE] A WIP to get JS tests running

thehenster:
The current problem is that JQuery can't be found.. A Slimmer thing?

The test is going to fail too but we're not getting to the assertion yet.


app/views/layouts/application.html.erb
```diff
@@ -6,12 +6,19 @@
   <!--[if IE 6]><%= stylesheet_link_tag "application-ie6" %><script>var ieVersion = 6;</script><![endif]-->
   <!--[if IE 7]><%= stylesheet_link_tag "application-ie7" %><script>var ieVersion = 7;</script><![endif]-->
   <!--[if IE 8]><%= stylesheet_link_tag "application-ie8" %><script>var ieVersion = 8;</script><![endif]-->
+
+  <%# We have to include jQuery here because it comes from slimmer %>
+  <%# which means we don't have it under test %>
+  <% if Rails.env.test? %>
+    <%= javascript_include_tag "jquery-2.2.3.min" %>
```
fofr: I don't like this approach, haven't thought of a better way yet. If doing this, it should match Slimmer as closely as possible, so:
* Javascript is loaded at the bottom of the page by slimmer.
* Slimmer uses jQuery version 1.11.3

You could pull the jQuery version from the frontend toolkit, that's what slimmer's version will do, via static.

PR #14, opened 2016-05-13
Adjust the position of the accordion icon

gemmaleigh:
The icon should be 16px in height, set it 50% from the top and with a
negative margin of half its height, so it always sits in the centre.

Also increase the spacing at the bottom of the header, so there isn’t a
“jump” in the position of the icon from plus to minus.

Screenshot before:

before

Screenshot after:

after


thehenster:
:+1:

PR #16, opened 2016-05-16
Save the state of the accordion using session storage

gemmaleigh:
When a section is opened - save its ID (with a prefix that identifies it as coming from the GOVUK Service Manual) to local storage with a value of 'Opened' and when it is closed, remove this key from session storage.

When the page loads, check session storage to see if any items have previously been stored and if so, open that section and set the `aria-expanded` attribute of the button to `true`.

cc @thehenster.

### DIFF
--- a/app/assets/javascripts/modules/accordion-with-descriptions.js
+++ b/app/assets/javascripts/modules/accordion-with-descriptions.js
@@ -16,25 +16,33 @@
       var totalSubsections = $element.find('.subsection__content').length;
 
       var $openOrCloseAllButton;
-      var GOVUKServiceManualTopic;
+      var GOVUKServiceManualTopic = createServiceManualTopicPrefix();
 
-      initialize();
+      addOpenCloseAllButton();
+      addButtonsToSubsections();
+      addIconsToSubsections();
+      addAriaControlsAttrForOpenCloseAllButton();
 
-      function initialize() {
-        addOpenCloseAllButton();
-        addSubsectionButtons();
-        addSubsectionIcons();
-        updateSubsectionControls();
-        getServiceManualTopicPrefix();
-        closeOpenSections();
-        checkSessionStorage();
+      closeOpenSections();
+      checkSessionStorage();
+
+      bindToggleForSubsections();
+      bindToggleOpenCloseAllButton();
+
+
+      function createServiceManualTopicPrefix() {
+        var GOVUKserviceManualPrefix = "GOVUK_service_manual";
+        var GOVUKserviceManualTopic = $('h1').text();
+        GOVUKserviceManualTopic = GOVUKserviceManualTopic.replace(/\s+/g,"_");
+        GOVUKserviceManualTopic = GOVUKserviceManualTopic.toLowerCase();
+        var GOVUKServiceManualTopic = GOVUKserviceManualPrefix+"_"+GOVUKserviceManualTopic+"_";
       }
 
       function addOpenCloseAllButton() {
         $element.prepend( '<div class="subsection-controls js-subsection-controls"><button aria-expanded="false">Open all</button></div>' );
       }
 
-      function addSubsectionButtons() {
+      function addButtonsToSubsections() {
         var $subsectionTitle = $element.find('.subsection__title');
 
         // Wrap each title in a button, with aria controls matching the ID of the subsection
@@ -43,11 +51,11 @@
         });
       }
 
-      function addSubsectionIcons() {
+      function addIconsToSubsections() {
         $subsectionHeader.append( '<span class="subsection__icon"></span>' );
       }
 
-      function updateSubsectionControls() {
+      function addAriaControlsAttrForOpenCloseAllButton() {
         // For each of the sections, create a string with all the subsection content IDs
         var ariaControlsValue = "";
         for (var i = 0; i < totalSubsections; i++) {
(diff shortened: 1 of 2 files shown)


gemmaleigh:
This needs tests, I'll add those.

app/assets/javascripts/modules/accordion-with-descriptions.js
```diff
@@ -122,6 +126,53 @@
         $node.attr("aria-expanded", state);
       }
 
+      function setSessionStorage() {
+        var isOpenSubsections = $('.subsection--is-open').length;
+        if (isOpenSubsections) {
+          var $openSubsections = $('.subsection--is-open');
+          $openSubsections.each(function(index) {
+            var subsectionOpenContentId = $(this).find('.subsection__content').attr('id');
+            sessionStorage.setItem( subsectionOpenContentId , subsectionOpenContentId);
```
thehenster: It looks like an example is a key of `subsection_content_1` with a value of `subsection_content_1`. And this session storage is shared across the www.gov.uk subdomain. Maybe a key of `service_manual_expand_topic_subsection_1` with a value of `true` will namespace it a little and make it a bit more explanatory if someone finds it and doesn't know what it is?

app/assets/javascripts/modules/accordion-with-descriptions.js
```diff
@@ -122,6 +126,53 @@
...
+          });
+        }
+      }
+
+      function removeSessionStorage() {
```
thehenster: I think I'd find this `removeSessionStorage()` and the above `setSessionStorage()` function names more representative of the role they have if they were called `setOpenTopicSections()` and `removeClosedTopicSections()`? Or `registerOpenTopicSections()` and `unregisterClosedTopicSections()`? Not sure.. Something like that maybe.

thehenster: Actually - ignore me. This is in the context of the accordion isn't it. I think it makes sense just like you have it.

app/assets/javascripts/modules/accordion-with-descriptions.js
```diff
@@ -122,6 +126,53 @@
...
+            sessionStorage.removeItem( subsectionClosedContentId , subsectionClosedContentId);
+          });
+        }
+      }
+
+      function checkSessionStorage() {
+
+        // Get section IDs from Session Storage
+        var sessionStorageItems = [];
+        for ( var i = 0, len = sessionStorage.length; i < len; ++i ) {
```
thehenster: Session storage is shared across subdomains so if any keys are set elsewhere on GOV.UK then they'll get returned here. Maybe we could iterate over the elements in the DOM and check if they exist in storage instead?

app/assets/javascripts/modules/accordion-with-descriptions.js
```diff
@@ -34,6 +34,17 @@
...
+      serviceManualTopic = serviceManualTopic.replace(/\s+/g,"_");
+      serviceManualTopic = serviceManualTopic.toLowerCase();
+      var GOVUKServiceManualTopic = serviceManualPrefix+"_"+serviceManualTopic;
+
+      // Add a data attribute to the wrapper with the current page's topic
+      var $subsectionWrapper = $element.find('.subsection-wrapper');
+      $subsectionWrapper.attr('data-service-manual-topic', GOVUKServiceManualTopic );
+
       // Get the open/close all button
       var $openOrCloseAllButton = $element.find('.js-subsection-controls button');
```
joelanman: Would allButtons plural be more readable/predictable?

joelanman: Sorry just realised that's not part of this pr

app/assets/javascripts/modules/accordion-with-descriptions.js
```diff
@@ -122,6 +144,59 @@
...
+          sessionStorageItems.push(sessionStorage.key( i ));
+        }
+
+        if (sessionStorageItems.length > 0) {
+          $.each( sessionStorageItems, function( key, value ) {
+            // console.log("key:"+key, "value:"+value);
+            // If any items in session storage start with GOVUK_Service_Manual and the topic name
+            if (value.indexOf(GOVUKServiceManualTopic) >= 0){
+              // console.log("items matched with prefix:"+GOVUKServiceManualTopic);
+              var subsectionContentId = value.toString();
```
joelanman: Is this needed? I think local storage can only store strings

app/assets/javascripts/modules/accordion-with-descriptions.js
```diff
@@ -34,6 +34,17 @@
         ariaControlsValue += "subsection_content_"+i+" ";
       }
 
+      // Create a unique prefix for each service manual topic
+      var serviceManualPrefix = "GOVUK_service_manual";
+      var serviceManualTopic = $('h1').text();
+      serviceManualTopic = serviceManualTopic.replace(/\s+/g,"_");
+      serviceManualTopic = serviceManualTopic.toLowerCase();
+      var GOVUKServiceManualTopic = serviceManualPrefix+"_"+serviceManualTopic;
```
joelanman: why add GOVUK prefix to the variable name at this point?

gemmaleigh: This variable is going to be saved in session storage, so it has been added to make it clear where these values have come from.

app/assets/javascripts/modules/accordion-with-descriptions.js
gemmaleigh: It's only one button :)

joelanman: ha ha! I read it wrong :)

app/assets/javascripts/modules/accordion-with-descriptions.js
```diff
@@ -8,51 +8,81 @@
...
+      initialize();
 
-      var $subsectionButton = $element.find('.subsection__button');
-      var $subsectionHeader = $element.find('.subsection__header');
+      function initialize() {
+        addOpenCloseAllButton();
+        addSubsectionButtons();
+        addSubsectionIcons();
+        updateSubsectionControls();
+        getServiceManualTopicPrefix();
```
thehenster: `getServiceManualTopicPrefix()` just returns a value so there shouldn't be a need to call it here.

thehenster: Ah - `GOVUKServiceManualTopic` is defined outside of the function `getServiceManualTopicPrefix()`. So there isn't any need for the return statement in `getServiceManualTopicPrefix()`.

PR #17, opened 2016-05-17
Filter search by service-manual content

AndrewVos:
This filters search results by "/service-manual" when a user searches
from the header inside a service manual content.

### DIFF
--- a/app/controllers/content_items_controller.rb
+++ b/app/controllers/content_items_controller.rb
@@ -1,9 +1,26 @@
 require 'gds_api/content_store'
+require 'slimmer/headers'
 
 class ContentItemsController < ApplicationController
+  include Slimmer::Headers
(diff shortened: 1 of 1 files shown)


app/controllers/content_items_controller.rb
```diff
@@ -1,9 +1,17 @@
...
 class ContentItemsController < ApplicationController
+  include Slimmer::Headers
   rescue_from GdsApi::HTTPForbidden, with: :error_403
 
   def show
+    set_slimmer_headers(
+      search_parameters: {
+        "filter_manual" => "/service-manual"
+      }.to_json,
+    )
```
thehenster: :clap: for working out how this works. Could you comment it a bit? It's a bit mind bending even when you know how it works..

AndrewVos: Sure thing sport

PR #19, opened 2016-05-19
Show Digital Service Manual as a header

AndrewVos:
image


thehenster:
I don't know what this is supposed to look like.. @gemmaleigh ^^

gemmaleigh:
Not like that.

We've got a problem as Slimmer strips out anything which isn't the propositional navigation.

We only need the proposition title, but there isn't an app with both that and the search box on the left.

http://alphagov.github.io/govuk_template/example-proposition-title.html

I think something like this would work, although we may well need to set a parent class on the header to adjust the styling as that's set in static.

I'm finding this impossible to preview locally. Any suggestions would be much appreciated.
    
   ```
 <header role="banner" id="global-header" class="service-manual-header">
      <div class="header-wrapper">
        <div class="header-global">
          <div class="header-logo">
            <a href="https://www.gov.uk" title="Go to the GOV.UK homepage" id="logo" class="content">
              <img src="https://assets.digital.cabinet-office.gov.uk/static/gov.uk_logotype_crown_invert_trans-78910ec0b2d46b9f4933556ba3d6863ca9bbcbefa8a4c811dce36f7026d04de8.png" width="35" height="31" alt=""> GOV.UK
            </a>
          </div>
          <div class='header-proposition'>
            <div class='content'>
              <a href='/govuk_template/' id='proposition-name'>Service Name</a>
            </div>
          </div>
          <a href="#search" class="search-toggle js-header-toggle">Search</a>
          <form id="search" class="site-search" action="/search" method="get" role="search">
            <div class="content">
              <label for="site-search-text">Search</label>
              <input type="search" name="q" id="site-search-text" title="Search" class="js-search-focus">
              <input class="submit" type="submit" value="Search">
            </div>
          </form>
        </div>
      </div>
    </header>


```

thehenster:
This is getting redesigned.

PR #20, opened 2016-05-20
Remove unused fields from service manual guide

thehenster:
Removes related_discussion which is no longer used and content_owner which has been moved to the links.

Related content schemas PR: https://github.com/alphagov/govuk-content-schemas/pull/313


dougdroper:
Looks good 👍

PR #21, opened 2016-05-20
Remove reference to FLAG_ENABLE_SERVICE_MANUAL flag

thehenster:
The service manual is enabled now. Also, this would have affected every test that ran after it.


gemmaleigh:
👍

PR #23, opened 2016-06-14
Base guide breadcrumb on service_manual_topics field

deborahchua:
This field was renamed from 'topics' as it is a reserved field.


thehenster:
LGTM

PR #24, opened 2016-06-16
Create two new functions to simplify creating a prefix

gemmaleigh:
- Get the h1
- Replace spaces with underscores
- Return a unique prefix for each subsection content area

cc @thehenster.


thehenster:
:+1:

PR #28, opened 2016-06-29
Make accordion description text darker

gemmaleigh:
The text will then default to $text-colour, which is #0b0c0c.

Before:
user research gov uk - before

After:
user research gov uk - after


36degrees:
👍

PR #29, opened 2016-06-29
Make page titles consistent

gemmaleigh:
Both Topic and Guide pages should display the page name, followed by -
Digital Service Manual - GOV.UK in the page title.

This was missing from the Topic page, and is added here.


36degrees:
👍

PR #30, opened 2016-06-29
Sort service standard points numerically

thehenster:
This fixes the bug where we sorted alphabetically and 10 came after 1 rather than 2.


36degrees:
LGTM 👍

PR #32, opened 2016-06-30
Remove unused ServiceManualGuidePresenter#related_discussion

thehenster:


36degrees:
Have verified that it's not called from anywhere. LGTM. 👍

PR #36, opened 2016-07-05
Add a section on the use of BEM to the README

gemmaleigh:
The Service manual frontend appplication is using the BEM methodology, with  modifications.
Add information about this to the README.

cc. @36degrees.


36degrees:
Perfick', thanks. 👍

PR #34, opened 2016-07-04
Use new page header module

36degrees:
### Topic
before - topic
### Guide
before - guide
### Service Standard
before - standard
### Point
before - point

### Topic

screen shot 2016-07-05 at 12 53 29

### Guide

screen shot 2016-07-05 at 12 53 37

### Service Standard

screen shot 2016-07-05 at 12 53 11

### Point

screen shot 2016-07-05 at 12 52 57

PR #37, opened 2016-07-06
Wraith Configuration and README change

36degrees:


thehenster:
Cool. I skimmed this was included but haven't tried running it.. Is it useful?

Why is the assets-origin.dev.gov.uk domain less resolvable in the VM than outside it?

36degrees:
Vagrant DNS creates a DNS resolver on your local machine, it doesn't do anything to help resolve inside the VM. It's a [known issue](https://github.com/BerlinVagrant/vagrant-dns#issues) listed in the plugin's documentation.

Having read about it in the docs, I went to use it to test the [page header module](https://github.com/alphagov/service-manual-frontend/pull/34) changes and discovered it wasn't configured sanely (still set up for government-frontend) and didn't work. So I thought I'd try and fix it.

36degrees:
For reference, I also created an issue in [government-frontend](https://github.com/alphagov/government-frontend/issues/162) so that the assets problem is documented for when/if they come up against it. They might have a better idea of how to make it work internally.

PR #39, opened 2016-07-12
Show the description for a point based on a boolean

thehenster:
Because we can use the description for the point text this bases whether we display the point text (or summary) on a boolean instead.

### DIFF
--- a/test/integration/service_manual_guide_test.rb
+++ b/test/integration/service_manual_guide_test.rb
@@ -37,7 +37,7 @@ class ServiceManualGuideTest < ActionDispatch::IntegrationTest
     end
   end
 
-  test "displays a summary if present" do
+  test "displays the description for a point" do
     setup_and_visit_example('service_manual_guide', 'point_page')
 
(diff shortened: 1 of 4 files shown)


test/integration/service_manual_guide_test.rb
```diff
@@ -37,15 +37,15 @@ class ServiceManualGuideTest < ActionDispatch::IntegrationTest
...
     within('.page-header__summary') do
       assert page.has_content?('Research to develop a deep knowledge of who the service users are')
     end
   end
 
-  test "the lede is not visible unless there is a summary" do
+  test "does not display the description for a normal guide" do
     setup_and_visit_example('service_manual_guide', 'service_manual_guide')
 
     refute page.has_css?('.lede')
```
36degrees: I *think* this might need updating since the header change went out. Might've missed it…

thehenster: Good spot - thanks

36degrees:
👍

PR #40, opened 2016-07-12
[DO NOT MERGE] A stab at the reduced header

thehenster:
We're trying to change the header in the following ways:

* Remove the government navigation links
* Scope the search to the service manual
* Change the search box label to 'Search the Service Manual'
* Add a title of 'Service Manual'

Here is the design:

searchbar results-02

This stab, iffy padding/styling aside, meets the criteria.

screen shot 2016-07-12 at 17 07 22

To achieve this we've:

* Set a slimmer header to remove the default search bar
* Stuck in our own `proposition-menu` that contains:
  * A proposition-name for the title
  * A custom search bar with the `filter-manual` parameter and our custom label

We almost got away with taking advantage of Slimmer's `Search-Parameters` header, which let us specify the `filter-manual` parameter, but as we needed the custom label it was removed.

Is there a better way?

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -20,3 +20,4 @@
 @import "modules/related-content";
 @import "modules/subsections";
 @import "modules/service-standard-point";
+@import "modules/search";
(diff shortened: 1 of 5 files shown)


thehenster:
This is getting redesigned. Keeping this one as a reference until the next one pops up.

PR #41, opened 2016-07-14
Don't render blank section headings for topics 'without sections'

36degrees:
Some topics have been configured with only one section which has neither a name nor a description. From a presentation point of view, we treat those topics as if they are 'sectionless'.

Previously the front-end did not handle that case very well – it still rendered a heading for the section, including a heading and description. This lead to blank h2s which is bad from an accessibility point of view (and also just not a great thing to do)

Now, for topics where there is exactly one section AND that section does not have a title, we skip the header and just output a list of links.

### DIFF
--- a/app/views/content_items/service_manual_topic.html.erb
+++ b/app/views/content_items/service_manual_topic.html.erb
@@ -22,35 +22,31 @@
 
 <div class="grid-row">
   <div class="column-two-thirds">
-    <% if @content_item.visually_collapsed? %>
-      <div class="js-hidden" data-module="accordion-with-descriptions">
-    <% else %>
-      <div>
-    <% end %>
-      <div class="subsection-wrapper">
-
-        <% @content_item.groups.each_with_index do |link_group, idx| %>
-
-          <div class="subsection">
-            <div class="subsection__header">
-              <h2 class="subsection__title"><%= link_group.name %></h2>
-              <p class="subsection__description"><%= link_group.description %></p>
+      <div<% if @content_item.visually_collapsed? %> class="js-hidden" data-module="accordion-with-descriptions"<% end %>>
+        <div class="subsection-wrapper">
+          <% @content_item.groups.each_with_index do |link_group, idx| %>
+            <div class="subsection">
+            <% if link_group.name.present? %>
+              <div class="subsection__header">
+                <h2 class="subsection__title"><%= link_group.name %></h2>
+              <% if link_group.description.present? %>
+                <p class="subsection__description"><%= link_group.description %></p>
+              <% end %>
+              </div>
+            <% end %>
+              <div class="subsection__content" id="subsection_content_<%= idx %>">
+                <ul class="subsection__list">
+                  <% link_group.linked_items.each do |linked_item| %>
+                    <li class="subsection__list-item">
+                      <%= link_to linked_item.label, linked_item.href %>
+                    </li>
+                  <% end %>
+                </ul>
+              </div>
             </div>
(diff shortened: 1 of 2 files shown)


36degrees:
Dependent on https://github.com/alphagov/govuk-content-schemas/pull/354

app/views/content_items/service_manual_topic.html.erb
```diff
@@ -22,35 +22,43 @@
...
+              <% link_group.linked_items.each do |linked_item| %>
+                <li class="subsection__list-item">
+                  <%= link_to linked_item.label, linked_item.href %>
+                </li>
+              <% end %>
+            </ul>
+          </div>
+        </div>
+      <% end %>
+    <% end %>
```
thehenster: Could you get away with putting the condition around just the heading bit?

```
<div class="grid-row">
  <div class="column-two-thirds">
    <div<% if @content_item.visually_collapsed? %> class="js-hidden" data-module="accordion-with-descriptions"<% end %>>
      <div class="subsection-wrapper">
        <% @content_item.groups.each_with_index do |link_group, idx| %>
          <div class="subsection">
            <% if @content_item.has_sections? %>
              <div class="subsection__header">
                <h2 class="subsection__title"><%= link_group.name %></h2>
                <p class="subsection__description"><%= link_group.description %></p>
              </div>
            <% end %>
            <div class="subsection__content" id="subsection_content_<%= idx %>">
              <ul class="subsection__list">
                <% link_group.linked_items.each do |linked_item| %>
                  <li class="subsection__list-item">
                    <%= link_to linked_item.label, linked_item.href %>
                  </li>
                <% end %>
              </ul>
            </div>
          </div>
        <% end %>
      </div>
    </div>
  </div>
```

I haven't tested ^^. It looks like we'd be adding in an unnecessary id in `<div class="subsection__content" id="subsection_content_<%= idx %>">` but that's pretty inoffensive.

If theres some unavoidable reason that we need to keep that if condition then we're probably in partial territory.

thehenster: Maybe the `.subsection-wrapper` is unneeded too? Does it cause an issue?

thehenster: I wonder if the thing that triggers this "should we display a heading or not" question is on a per-group level rather than at the whole-topic level.

How about..

```
<div class="grid-row">
  <div class="column-two-thirds">
    <div<% if @content_item.visually_collapsed? %> class="js-hidden" data-module="accordion-with-descriptions"<% end %>>
      <div class="subsection-wrapper">
        <% @content_item.groups.each_with_index do |link_group, idx| %>
          <div class="subsection">
            <% if link_group.name.present? %>
              <div class="subsection__header">
                <h2 class="subsection__title"><%= link_group.name %></h2>
                <p class="subsection__description"><%= link_group.description %></p>
              </div>
            <% end %>
            <div class="subsection__content" id="subsection_content_<%= idx %>">
              <ul class="subsection__list">
                <% link_group.linked_items.each do |linked_item| %>
                  <li class="subsection__list-item">
                    <%= link_to linked_item.label, linked_item.href %>
                  </li>
                <% end %>
              </ul>
            </div>
          </div>
        <% end %>
      </div>
    </div>
  </div>
```

And it's up to the publisher to make sure there isn't a weird mix of titles and no titles?

36degrees:
Did this again in a much simpler manner as per comments. Thoughts?

36degrees:
[Ignoring whitespace on the diff](https://github.com/alphagov/service-manual-frontend/pull/41/files?w=0) makes it clearer what's going on.

PR #42, opened 2016-07-15
Display public_updated_at to user for guides

thehenster:
https://trello.com/c/7bdchGKi/383-a-guide-s-last-updated-time-updates-each-time-a-topic-is-published

The public_updated_at time is the user facing timestamp to hopefully represent the last major published at time. The updated_at time, which we were using, is updated if a related thing, a topic in this case, updates it.

Move the use of time_ago_in_words into the view and test at the integration level. Testing the rails helper at the unit didn't provide much more confidence than the framework tests give us and this way we can check the timestamp is actually on the page.


36degrees:
👍

PR #44, opened 2016-08-03
Display change history on guides

36degrees:
screen shot 2016-08-05 at 15 38 17
screen shot 2016-08-05 at 15 39 03
screen shot 2016-08-05 at 15 38 59

Waiting for https://github.com/alphagov/govuk-content-schemas/pull/369 to be merged before these tests will pass.


test/integration/service_manual_guide_test.rb
```diff
@@ -66,4 +66,31 @@ class ServiceManualGuideTest < ActionDispatch::IntegrationTest
...
+      assert page.has_content? 'Guidance created'
+    end
+  end
+
+  test 'omits the previous history if there is only one change' do
+    setup_and_visit_example('service_manual_guide', 'service_manual_guide')
+
+    refute page.has_content? 'Show all page updates'
+    refute page.has_css? '.change-history__past'
+  end
```
thehenster: We've covered off the error in the view, but I'd find it useful to have the test for the scenario in which the content_history is empty and we display neither the previous changes or the most recent change.

The absence of a major update will be increasingly odd over time so having the test is a kind "yeah, we know and this is the reason" will be useful.

app/views/content_items/service_manual_guide.html.erb
```diff
@@ -78,5 +78,53 @@
...
+    <div class="change-history" data-module="toggle">
+      <div class="change-history__latest-change">
+        <dl>
+        <% if @content_item.content_owners.any? %>
+          <dt>Published by:</dt>
+          <% @content_item.content_owners.each do |content_owner| %>
+            <dd>
+              <%= link_to content_owner.title, content_owner.href %>
+            </dd>
+          <% end %>
```
thehenster: Could the 'Published by' bit be outside the context of the change history? It feels to me like we'd have that even if we had the change history or not.

app/views/content_items/service_manual_guide.html.erb
```diff
@@ -78,5 +78,53 @@
...
+        <% end %>
+        <% if @content_item.latest_change.present? %>
+          <dt>Last update:</dt>
+          <dd>
+            <% time = @content_item.latest_change.public_timestamp %>
+            <time datetime="<%= time.iso8601 %>" class="timestamp"><%= time.strftime("%e %B %Y") %></time>
```
thehenster: :clap: for the use of the time element.

To get that strftime out of the view, this is the Rails-y way to do it: http://guides.rubyonrails.org/i18n.html#adding-date-time-formats

app/views/content_items/service_manual_guide.html.erb
```diff
@@ -78,5 +78,53 @@
...
+        <li>
+          <% time = previous_change.public_timestamp %>
+          <time datetime="<%= time.iso8601 %>" class="timestamp"><%= time.strftime("%e %B %Y") %></time>
+          <p class="change-history__summary">
+            <%= previous_change.note %>
+          </p>
+
+          <div class="change-history__reason">
+            <%= simple_format(previous_change.reason_for_change) %>
+          </div>
```
thehenster: Maybe it would require a play to see if it aids readability rather than hinders it.. But this could go in a `_change_history.html.erb` partial.

app/views/content_items/service_manual_guide.html.erb
thehenster: Mentioned IRL but this could be partial'ed if it actually makes it more readable.

thehenster:
Build failure but it's looking good to me.

36degrees:
If you're happy to merge https://github.com/alphagov/govuk-content-schemas/pull/371 I think this'll start passing again.

PR #45, opened 2016-08-12
Upgrade to rails 4 2

thehenster:
This just so happens to be mergeable.. But master isn't deployable at the moment.


36degrees:
👍

PR #48, opened 2016-08-19
Include a link to the email alert signup page on the topic page

36degrees:
As the topics are republished they'll be associated by link to an email_alert_signup for the topic. If that link is present, include it as part of the topic page.

Expecting tests to fail until related changes in govuk-content-schemas are merged.


36degrees:
Would definitely like to get some feedback on the frontend changes here. I feel like .email-link being a module is overkill, but it doesn't seem to belong anywhere else. Also not sure on the use of `aside` – should there be an `aside` for each thing, or is it OK to have one `aside` with multiple different things in it as I've done here?

Perhaps @gemmaleigh could take a look?

test/integration/service_manual_topic_test.rb
```diff
@@ -40,6 +40,13 @@ class ServiceManualTopicTest < ActionDispatch::IntegrationTest
     end
   end
 
+  test "it includes a link to subscribe for email alerts" do
+    setup_and_visit_example('service_manual_topic', 'service_manual_topic')
+
+    assert page.has_link?("email",
+      href: "http://www.dev.gov.uk/service-manual/test-expanded-topic/email-signup")
```
thehenster: A minor thing. Matching just the path will make this slightly more robust. You can use a regex.

thehenster:
What is a block and an element in BEM is definitely an art rather than a science. In that case I might use `email-subscription__link` in the file email-subscription.scss. Or it could just about live in topic.scss as `topic__subscribe` but you've decided against that?

One of the big goals of the blocks, like modules, is reusability. Once we've stuck these links on the homepage and the standard the reusable part will probably be obvious. So I wouldn't worry about it to much.. we'll get it right then.

Dunno about the `aside`.

test/integration/service_manual_topic_test.rb
36degrees: Would it in fact just make more sense to use the `base_path` rather than the `web_url` to link here?

36degrees:
@thehenster Are you happy with this now? I've changed the presenter to use (and test for) the base path rather than providing a full url.

gemmaleigh:
Would it be possible to move the `.email-link` scss into the related-item component for now? 

It isn't used anywhere else.. so we could have `.related-item__email-link` and if it is reused out of this context, we can look to move it at a later date.

36degrees:
@gemmaleigh Done. 👍

PR #49, opened 2016-08-23
Timestamp for first draft

thehenster:
https://trello.com/c/DK1xNwvb

The `public_updated_at` timestamp is not available when previewing the first edition. To avoid a bug we fall back on the `updated_at` field which represents the last time the content item was updated _for any reason_ in the content store. In between a user saving a draft and previewing it, this is most likely going to be the time they saved the draft.

If we ever want to get more precise about this timestamp we should start sending `last_edited_at` to the publishing-api.

I've added a test helper for the draft stack examples which is a lot of trumpeting for deleting a single key from a hash. However, organisation wide I think it's still fairly unknown behaviour so it's a good thing to be a little shouty about.

### DIFF
--- a/test/integration/service_manual_guide_test.rb
+++ b/test/integration/service_manual_guide_test.rb
@@ -1,6 +1,27 @@
 require 'test_helper'
 
 class ServiceManualGuideTest < ActionDispatch::IntegrationTest
+  test "shows the time it was saved and it hasn't been published yet" do
+    now = "2015-10-10T09:00:00+00:00"
+    last_saved_at = "2015-10-10T08:55:00+00:00"
+
+    travel_to(now) do
+      example = get_content_example_by_format_and_name(
+        'service_manual_guide',
+        'service_manual_guide',
+        'updated_at' => last_saved_at
+      )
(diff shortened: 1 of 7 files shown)


test/integration/service_manual_guide_test.rb
```diff
@@ -1,6 +1,28 @@
 require 'test_helper'
 
 class ServiceManualGuideTest < ActionDispatch::IntegrationTest
+  test "shows the time it was saved and hasn't been published yet" do
```
36degrees: Is this description correct?

thehenster: It's not wrong.. but it could be clearer. Does "before being published, it displays the time it was saved" read better?

36degrees: Yeah, it's the use of the word 'and' makes it sound like you're _testing_ that it hasn't been published yet. `shows the time it was saved if it hasn't been published yet` would also work.

36degrees:
Minor comment about a test description that doesn't quite work, but happy for this to be merged when that's fixed if you want to do anything about it.

PR #50, opened 2016-08-23
Build the service standard points from the expanded links

thehenster:
Use link expansion to avoid saving and publishing the service standard every time a point changes.

Publisher PR: https://github.com/alphagov/service-manual-publisher/pull/341


36degrees:
You on the test failures here?

36degrees:
Aah, wait, requires the updated examples in schemas by any chance?

36degrees:
👍

PR #51, opened 2016-08-25
[For discussion] Rename modules add utilities

gemmaleigh:
This isn't ready to merge yet, but since there were quite a few modules which didn't do anything other than add top or bottom spacing:
- govspeak-wrapper
- community-contact
- main-service-manual

These have been replaced with utility classes for vertical spacing.

I've removed the typography module as it would be better to use the typography mixins to set styling for elements within components, rather than everywhere.

This also fixes the current issue with the page-header component, where margins are inherited from the paragraph styles added by typography.scss, and fixes the alignment of the 'Feedback' link.

Before:
agile and government services an introduction digital service manual gov uk

After:
agile and government services an introduction digital service manual gov uk


app/assets/stylesheets/application.scss
```diff
@@ -1,25 +1,22 @@
...
-@import "mixins/margins";
+// Components
+@import "components/change-history";
+@import "components/metadata";
+@import "components/notice";
+@import "components/page-contents";
+@import "components/page-header";
+@import "components/related-content";
+@import "components/service-standard-point";
+@import "components/subsections";
```
thehenster: Do you think there is any risk of these being confused with "GOV.UK components"?

gemmaleigh: Hadn't considered that, good point.

We could use `/service-manual-components/` to make it clear they are components for this project.

app/views/content_items/service_manual_guide.html.erb
```diff
@@ -1,4 +1,4 @@
-<%= content_for :page_class, "service-manual" %>
+<%= content_for :page_class, "u-margin-bottom-large" %>
```
thehenster: Are we getting much benefit from using a `content_for` for this? The only thing that isn't in this template (at the moment) is the phase label, which I don't think needs to be inside `.u-margin-bottom-large`?

I thought the intent of this thing was a kind of convention to allow us to apply custom styles to a particular page type.

Could we bin it?

app/assets/stylesheets/application.scss
thehenster: Yeah - that'd be good :+1:

PR #56, opened 2016-09-14
Remove search on homepage

36degrees:
This branch is built off of `homepage-intro-and-search`. Once that branch has merged, the target for this branch can be updated to `homepage`.

### DIFF
--- a/app/controllers/content_items_controller.rb
+++ b/app/controllers/content_items_controller.rb
@@ -6,26 +6,13 @@ class ContentItemsController < ApplicationController
   rescue_from GdsApi::HTTPForbidden, with: :error_403
 
   def show
-    # The Slimmer middleware is responsible for intercepting the response and inserting the navigation header into the HTML body. We need to set the request headers here for the search form in the header to have the correct hidden input to scope the search results to just the service manual.
-
-    # It should look something like:
-
-    # <form id="search" class="site-search" action="/search" method="get" role="search">
-    #   <input type="search" name="q" id="site-search-text" title="Search" class="js-search-focus">
-    #   <input class="submit" type="submit" value="Search">
-    #   <input type="hidden" name="filter_manual" value="/service-manual">
-    # </form>
-    set_slimmer_headers(
-      search_parameters: {
-        "filter_manual" => "/service-manual"
-      }.to_json,
-    )
-
     if load_content_item
       set_expiry
       set_locale
+      configure_header_search
       render content_item_template
     else
+      configure_header_search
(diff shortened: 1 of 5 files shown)


app/controllers/content_items_controller.rb
```diff
@@ -67,7 +54,28 @@ def content_store
...
+    if @content_item.present? && @content_item.is_homepage?
+      remove_header_search
+    else
+      scope_header_search_to_service_manual
+    end
+  end
+
+  def scope_header_search_to_service_manual
+    # Slimmer is middleware which wraps the service manual in the GOV.UK header
+    # and footer. We set a request header so that Slimmer adds a hidden field to
```
thehenster: I think we're setting a response header.

app/controllers/content_items_controller.rb
```diff
@@ -67,7 +54,28 @@ def content_store
...
+    set_slimmer_headers(
+      search_parameters: {
+        "filter_manual" => "/service-manual"
+      }.to_json,
+    )
+  end
+
+  def remove_header_search
+    set_slimmer_headers(remove_search: true)
+  end
```
thehenster: We could push these headers (both the search_parameters and remove_search) into the presenters. Maybe a `#slimmer_headers` method on the presenter classes? The benefits would be that we don't have to add the `#is_homepage?` query method because it knows what itself is, and it would remove the conditional out of the controller which is always nice.

36degrees: Not sure about this. It seems like a weird concern for the presenter to have. But it would simplify things a bit. It's not where I'd expect to find response headers being set though…?

app/controllers/content_items_controller.rb
```diff
@@ -6,26 +6,13 @@ class ContentItemsController < ApplicationController
...
-    set_slimmer_headers(
-      search_parameters: {
-        "filter_manual" => "/service-manual"
-      }.to_json,
-    )
-
     if load_content_item
       set_expiry
       set_locale
+      configure_header_search
```
mcmillan: Does this have to happen after `set_expiry` and `set_locale`? If not, you could DRY it up a bit put `configure_header_search` outside the conditional here.

36degrees: It needs to happen after load_content_item. So it could be

```ruby
content = load_content_item
configure_header_search

if content
  [...]
```

if you think that's better?

mcmillan: Ah, yeah – sorry. I think the current way is more concise. Ignore me!

PR #57, opened 2016-09-15
Improve this page

thehenster:
https://trello.com/c/HN1L3EfK

# Why

The service manual wants a way to allow users to give positive feedback, as well as negative.

# What

A widget to offer a suggestion on how to improve the page or to feedback that the page was useful. This is an alternative to the existing 'Report a Problem' widget that gives encourages positive feedback.

It works by submitting an AJAX request to the feedback frontend, which in turn submits a request to the support-api, which in turn submits the feedback to Zendesk.

Custom event tracking via Google Analytics is supported if the GOV.UK Analytics module is loaded.

The fallback behaviour in the absense of javascript is to send the user to the GOV.UK contact page.

## A widget just above the footer
screen shot 2016-09-16 at 15 55 57

## When you click on "No" or "Is there anything wrong with this page?"
screen shot 2016-09-16 at 15 56 16

## Success
screen shot 2016-09-16 at 15 56 37

# Other moving parts

* The feedback frontend acts as a publicaly accessible proxy inbetween the user's browser and the support-api https://github.com/alphagov/feedback/pull/177
* The support-api receives the request from the feedback frontend and creates a Zendesk ticket https://github.com/alphagov/support-api/pull/96
* gds-api-adapters changed to add helpers and test helpers for the new endpoint https://github.com/alphagov/gds-api-adapters/pull/576


app/assets/javascripts/modules/improve-this-page.js
```diff
@@ -0,0 +1,160 @@
...
+        action: that.pageIsUsefulButton.data('track-action')
+      }
+    }
+
+    this.renderErrors = function (errors) {
+      that.feedbackForm.find('.js-error').remove();
+
+      $.each(errors, function (attrib, messages) {
+        $.each(messages, function (index, message) {
+          var errorNode = $('<div class="improve-this-page__error js-error">' + attrib + ' ' + message + '.</div>');
```
36degrees: I'd be tempted to use `$('<div…>').text(attrib + ' ' + message)` here to escape `attrib` and`message`.

app/assets/javascripts/modules/improve-this-page.js
```diff
@@ -0,0 +1,160 @@
...
+          var field = that.feedbackForm.find('[name="'+ attrib + '"]');
+
+          // If a field with the same name as the error attribute then display
```
36degrees: Think you're missing a word here.

app/assets/javascripts/modules/improve-this-page.js
```diff
@@ -0,0 +1,160 @@
...
+      this.controller = new Controller(this.view);
+
+      this.controller.init();
+    }
+  };
+
+  function View ($element) {
+    var that = this;
+
+    this.pageIsUsefulButton = $element.find('.js-page-is-useful');
```
36degrees: Do we prefix the names of jquery objects with `$` here, or did I imagine that? There's a number of places throughout this file where that would apply if so.

app/assets/javascripts/modules/improve-this-page.js
```diff
@@ -0,0 +1,160 @@
...
+      $.each(errors, function (attrib, messages) {
+        $.each(messages, function (index, message) {
+          var errorNode = $('<div class="improve-this-page__error js-error">' + attrib + ' ' + message + '.</div>');
+          var field = that.feedbackForm.find('[name="'+ attrib + '"]');
+
+          // If a field with the same name as the error attribute then display
+          // the error inline with the field. If a matching field doesn't
+          // exist then display it above the form.
+          if (field.length) {
+            field.before(errorNode);
```
36degrees: Is there any thing to clear existing validation messages so that multiple failures don't get duplicated?

app/views/layouts/application.html.erb
```diff
@@ -23,6 +23,45 @@
...
+      </div>
+      <div class="js-feedback-form js-hidden">
+        <div class="grid-row">
+          <div class="column-two-thirds">
+            <div class="js-errors"></div>
+            <form>
+              <input type="hidden" name="path" value="<%= request.original_url -%>">
+              <div class="improve-this-page__section">
+                <label>How should we improve this page?</label>
+                <textarea name="description" rows="5"></textarea>
```
36degrees: Worth adding the `required` flag for inline browser validation?

spec/javascripts/helpers/with_govuk_analytics.js
```diff
@@ -0,0 +1,14 @@
+// Temporarily overrides GOVUK.analytics and either returns it
```
36degrees: Unfinished comment?

app/assets/javascripts/modules/improve-this-page.js
```diff
@@ -0,0 +1,160 @@
...
+  function View ($element) {
+    var that = this;
+
+    this.pageIsUsefulButton = $element.find('.js-page-is-useful');
+    this.offerFeedbackButton = $element.find('.js-offer-feedback');
+    this.feedbackForm = $element.find('.js-feedback-form');
+    this.submitFeedbackButton = that.feedbackForm.find('[type=submit]');
+    this.prompt = $element.find('.js-prompt');
+
+    this.onPageIsUsefulButtonClicked = function (callback) {
```
36degrees: I'm assuming this is how we do things but I find this really hard to follow… when a thing is clicked, click the thing? What?

app/assets/javascripts/modules/improve-this-page.js
thehenster: Yeah - we do. I have missed some. I'll update them.

thehenster: There were quite a few in fact! Updated.

app/assets/javascripts/modules/improve-this-page.js
thehenster: I've tried to keep the jQuery interacting with the DOM bits separate from the orchestration bit.

It's the idea behind the segregated DOM which I think is cool. http://radar.oreilly.com/2014/01/keeping-jquery-in-check.html http://martinfowler.com/bliki/SegregatedDOM.html

app/assets/javascripts/modules/improve-this-page.js
thehenster: Yeah, that's better. Thanks.

app/views/shared/_improve_this_page.html.erb
```diff
@@ -0,0 +1,38 @@
...
+        <form>
+          <input type="hidden" name="path" value="<%= request.original_url -%>">
+          <div class="improve-this-page__section">
+            <label>How should we improve this page?</label>
+            <textarea name="description" rows="5"></textarea>
+          </div>
+          <p>Include your name and email address if you'd like us to get back to you.</p>
+          <div class="improve-this-page__section">
+            <label>Name</label>
+            <span>(optional)</span>
```
thehenster: The whole thing is busted for accessibility. I'm hoping to have a chat with Gemma about it.

These labels will need `for` attributes. The yes and no links beside "Is this page useful?" will probably just read out "yes" and "no" and I don't have a good idea how to go about that one. Errors popping up via AJAX sound a little intimidating too, although using the browser validation for the description field will probably be as good as it can be and cover most cases this form is likely to hit.

Redirecting to the contact page, like the JS fallback does, is an option too but I'd much prefer not to.

app/views/layouts/application.html.erb
thehenster: Good shout, it would be worth adding. I suspect the way I'm submitting the data by plucking out the field values out with `.serialize()` will circumvent the validation too. Hmm.

thehenster: I've added commits to bind to the submit event on the form and have added the required attribute. I'm going to leave the code in which renders the errors inline. It's only 3-4 lines and a test which doesn't seem like much of a cost for not having to re-add it if we add other validations to the support-api.

app/assets/javascripts/modules/improve-this-page.js
```diff
@@ -0,0 +1,164 @@
...
+        action: that.$pageIsUsefulButton.data('track-action')
+      }
+    }
+
+    this.renderErrors = function (errors) {
+      that.$feedbackFormContainer.find('.js-error').remove();
+
+      $.each(errors, function (attrib, messages) {
+        $.each(messages, function (index, message) {
+          var $errorNode = $('<div/>)', {
```
36degrees: Is that closing parenthesis after `<div/>` meant to be there?

app/views/shared/_improve_this_page.html.erb
```diff
@@ -0,0 +1,41 @@
...
+      Is there anything wrong with this page? <span class="visually-hidden"> this page is not useful</span>
+    <% end %>
+  </div>
+  <div class="js-feedback-form js-hidden">
+    <div class="grid-row">
+      <div class="column-two-thirds">
+        <div class="js-errors"></div>
+        <form>
+          <input type="hidden" name="path" value="<%= request.original_url -%>">
+          <input type="hidden" name="user_agent" value="<%= request.user_agent -%>">
```
36degrees: I need to better understand what ERB escapes when, so forgive me if this is a stupid question – if the user agent or the original url contain quotes or other special characters I'm presuming these'll be escaped as HTML entities?

app/views/shared/_improve_this_page.html.erb
```diff
@@ -0,0 +1,41 @@
...
+          <div class="improve-this-page__section">
+            <label for="improve-this-page_description">How should we improve this page? <span class="visually-hidden">required</span></label>
+            <textarea name="description" rows="5" required id="improve-this-page_description"></textarea>
```
36degrees: To avoid confusion between BEM classes and IDs, is it worth just using a hyphen here (`improve-this-page-description`)? It can be a bit hard to parse the classes and IDs in this block as they're so similar.

(Alternative suggestion – can the association between label and input be made implicit by wrapping the input with the label? Then the `for` and `id` attributes can be removed)

app/views/shared/_improve_this_page.html.erb
```diff
@@ -0,0 +1,41 @@
...
+          </div>
+          <p>Include your name and email address if you'd like us to get back to you.</p>
+          <div class="improve-this-page__section">
+            <label  for="improve-this-page_name">Name <span class="improve-this-page__hint">(optional)</span></label>
+            <input type="text" name="name" id="improve-this-page_name">
+          </div>
+          <div class="improve-this-page__section">
+            <label  for="improve-this-page_email">Email <span class="improve-this-page__hint">(optional)</span></label>
```
36degrees:
```
<label  for="
       ^ 👾 
```

36degrees: And on :27 too.

test/integration/improve_this_page_test.rb
```diff
@@ -0,0 +1,20 @@
+require 'test_helper'
+
+class ImproveThisPageTest < ActionDispatch::IntegrationTest
```
36degrees: Any plans to test when javascript is enabled? Even if only that clicking the feedback link makes the relevant form / message visible?

app/assets/javascripts/modules/improve-this-page.js
thehenster: I got it from an example in the jQuery docs. Search for "Create some DOM elements" on http://api.jquery.com/jquery/

app/views/shared/_improve_this_page.html.erb
thehenster: Yeah, they should do, unless Rails has made those strings "safe", which I doubt it has. I think it's OK, even desirable, having those values HTML escaped isn't it? If someone has put a script tag in their user agent I quite fancy escaping it.. Ampersands will be escaped in a query string in the URL but we don't have paths based on query strings.. I'm not feeling a lot of urgency to make a change here. Any feelings?

app/assets/javascripts/modules/improve-this-page.js
36degrees: Yeah, but that still doesn't have the `)` I'm talking about…

app/views/shared/_improve_this_page.html.erb
36degrees: Yes, I was checking that it _was_ escaped. Not suggesting that it was a problem were that the case.

test/integration/improve_this_page_test.rb
thehenster: I definitely wrote that test once.. I've lost it somehow. I agree it's needed. Ta.

app/assets/javascripts/modules/improve-this-page.js
thehenster: Thanks! I saw it in the end :)

app/views/shared/_improve_this_page.html.erb
thehenster: :+1:

test/integration/improve_this_page_test.rb
```diff
@@ -17,4 +17,22 @@ class ImproveThisPageTest < ActionDispatch::IntegrationTest
...
+      within('.improve-this-page') do
+        page.click_link 'No'
+      end
+
+      refute feedback_form_is_hidden
+    end
+  end
+
+  def feedback_form_is_hidden
+    page.find('.js-feedback-form')[:class].include?('js-hidden')
```
36degrees: I think Capybara already has methods to assert that an element is *actually* visible on the page, rather than tying this to a particular class?

thehenster: It does.. but they don't work when the CSS that does the hiding isn't available inside the test. I think this is a lesser evil than running a service or "stubbing out" parts of govuk_template.

app/views/shared/_improve_this_page.html.erb
thehenster: I should almost definitely be embarrassed. I didn't know that was valid. It works well with the screenreader too!

test/integration/improve_this_page_test.rb
36degrees: Aaaah of course. OK. Very good. You may continue. Proceed to Go. Collect $200.

36degrees: Actually, may be worth a comment. ¯\\_(ツ)_/¯

thehenster: Sure! Added.

spec/javascripts/improve-this-page-spec.js
```diff
@@ -0,0 +1,207 @@
...
+          '<a href="/contact/govuk" class="js-offer-feedback">Is there anything wrong with this page?</a>' +
+        '</div>' +
+      '</div>' +
+      '<div class="js-feedback-form js-hidden" data-track-category="improve-this-page" data-track-action="give-feedback">' +
+        '<div class="js-errors"></div>' +
+        '<form>' +
+          '<input type="hidden" name="url" value="http://example.com/path/to/page"></input>' +
+          '<input type="hidden" name="user_agent" value="Safari"></input>' +
+          '<div>' +
+            '<label>How should we improve this page?</label>' +
```
gemmaleigh: The markup in the fixture doesn't match the markup you are using. 
You've got labels wrapping the label text, hint text and input field - but not here.

thehenster: Thanks for this - I've added a commit.

In the past I've managed to get javascript tests loading the rails partial as a fixture which avoids this kind of problem. I'll have a play in a separate branch.

PR #58, opened 2016-09-16
Add meta descriptions

rubenarakelyan:
The meta description tag is used for the index page of the service manual but none of the sub-pages. This commit adds meta description tags to all pages, the contents of which are set to the description associated with the relevant part of the service manual. This adds descriptions to external search engine results when the pages are re-indexed.

Trello: https://trello.com/c/bbAwFAQ1/108-add-meta-descriptions-to-pages-missing-them


36degrees:
Thanks, @rubenarakelyan

PR #59, opened 2016-09-21
[Do not merge, demo only] Add sticky nav

gemmaleigh:
This is a branch to demo [GOVUK.StickAtTopWhenScrolling](https://github.com/alphagov/govuk_frontend_toolkit/blob/master/javascripts/govuk/stick-at-top-when-scrolling.js).

Steps for this PR:

- [ ] Merge the changes to GOVUK.StickAtTopWhenScrolling in the govuk_frontend_toolkit
- [ ] Bump the version of the govuk_frontend_toolkit
- [ ] Update the version of the govuk_frontend_toolkit that service-manual-frontend is using
- [ ] Remove the duplicate stick-at-top-when-scrolling.js copied to this repo in `app/assets/javascripts`

Closing in favour of #61.

### DIFF
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -5,4 +5,4 @@
 //= require_tree .
 
 window.GOVUK.stickAtTopWhenScrolling.init();
-window.GOVUK.stopScrollingAtFooter.init();
+window.GOVUK.stopScrollingAtFooter.addEl($('.page-contents'));
--- a/app/assets/javascripts/application.js
+++ b/app/assets/javascripts/application.js
@@ -1,7 +1,6 @@
 // from govuk_frontend_toolkit
-//= require govuk/stick-at-top-when-scrolling
+//= require stick-at-top-when-scrolling
 //= require govuk/stop-scrolling-at-footer
-
 //= require_tree .
 
 window.GOVUK.stickAtTopWhenScrolling.init();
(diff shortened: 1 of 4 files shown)


app/assets/stylesheets/modules/_page-contents.scss
```diff
@@ -3,6 +3,7 @@
 
   @include media(tablet) {
     margin-top: 2em;
+    padding-bottom: 2em;
```
thehenster: Is the plan to move this to govuk_frontend_toolkit to?

app/assets/javascripts/application.js
```diff
@@ -5,4 +5,4 @@
 //= require_tree .
 
 window.GOVUK.stickAtTopWhenScrolling.init();
-window.GOVUK.stopScrollingAtFooter.init();
+window.GOVUK.stopScrollingAtFooter.addEl($('.page-contents'));
```
thehenster: I wrote the below comment, then looked at the PR on govuk_frontend_toolkit, and my comment isn't related to what you're changing. So.. ignore this.. but maybe for one day in the future.

I like the passing of the element into `addEl()`. How about doing something similar with the `init()` above?

eg.

```
window.GOVUK.stickAtTopWhenScrolling.init($(".js-stick-at-top-when-scrolling"));
```

Or.. Could the `data-module` technique be used?

PR #61, opened 2016-09-30
Highlight anchors in the page contents navigation, on scroll

gemmaleigh:
This PR adds a HighlightActiveSectionHeading module.

This will highlight an anchor link in the "Page contents" left hand navigation, when a user has scrolled past the corresponding heading in the main content area. 

This PR also uses the GOVUK.StickAtTopWhenScrolling JS, to "stick" the navigation, once the user has scrolled past the left hand page contents and to "unstick" for smaller viewports.

Steps for this PR:
- [x] Merge the changes to GOVUK.StickAtTopWhenScrolling in the govuk_frontend_toolkit
- [x] Bump the version of the govuk_frontend_toolkit
- [x] Update the version of the govuk_frontend_toolkit that service-manual-frontend is using 
- [x] Remove commits: 
45ebd8b - duplicating duplicate the GOVUK.StickAtTopWhenScrolling() JS 
and 
a5f8f41 - revert to use govuk/stick-at-top rather than temp/stick-at-top

### DIFF
--- a/app/assets/javascripts/modules/highlight-active-section-heading.js
+++ b/app/assets/javascripts/modules/highlight-active-section-heading.js
@@ -1,0 +1,144 @@
+;(function (Modules, root) {
+  'use strict'
+
+  var $ = root.$
+  var $window = $(root)
+
+  Modules.HighlightActiveSectionHeading = function () {
+    var self = this
+    var _hasResized = true
+    var _hasScrolled = true
+    var _interval = 50
+    var anchorIDs = []
+
+    self.getWindowDimensions = function () {
+      return {
+        height: $window.height(),
+        width: $window.width()
+      }
+    }
+
+    self.getWindowPositions = function () {
+      return {
+        scrollTop: $window.scrollTop()
+      }
+    }
+
+    self.getElementOffset = function ($el) {
+      return $el.offset()
+    }
+
+    self.start = function ($el) {
+      $window.resize(self.hasResized)
+      $window.scroll(self.hasScrolled)
+
+      setInterval(self.checkResize, _interval)
+      setInterval(self.checkScroll, _interval)
+
+      self.$anchors = $el.find('.js-page-contents a')
+      self.getAnchors()
+
+      self.checkResize()
+      self.checkScroll()
+    }
+
+    self.hasResized = function () {
+      _hasResized = true
+      return _hasResized
+    }
+
+    self.hasScrolled = function () {
+      _hasScrolled = true
+      return _hasScrolled
+    }
+
+    self.checkResize = function () {
+      if (_hasResized) {
+        _hasResized = false
+        _hasScrolled = true
+      }
+    }
+
+    self.checkScroll = function () {
+      if (_hasScrolled) {
+        _hasScrolled = false
+        var windowDimensions = self.getWindowDimensions()
+        if ( windowDimensions.width <= 768) {
+          self.removeActiveItem()
+        } else {
+          self.updateActiveNavItem()
+        }
+      }
+    }
+
+    self.getAnchors = function () {
+      $.each(self.$anchors, function(i) {
+        var anchorID = $(this).attr('href')
+        // e.g. anchorIDs['#meeting-the-digital-service-standard', '#understand-your-users', '#research-continually']
+        anchorIDs.push(anchorID)
+      })
+    }
+
+    self.getHeadingPosition = function ($theID) {
+      return $theID.offset()
+    }
+
+    self.getNextHeadingPosition = function ($theNextID) {
+      return $theNextID.offset()
+    }
+
+    self.getFooterPosition = function ($theID) {
+      return $theID.offset()
+    }
+
+    self.getDistanceBetweenHeadings = function (headingPosition, nextHeadingPosition) {
+      var distanceBetweenHeadings = (nextHeadingPosition - headingPosition)
+      return distanceBetweenHeadings
+    }
+
+    self.updateActiveNavItem = function () {
+      var windowVerticalPosition = self.getWindowPositions().scrollTop
+      var footerPosition = self.getFooterPosition($('#footer'))
+
+      $.each(self.$anchors, function(i) {
+
+        var theID = anchorIDs[i]
+        var theNextID = anchorIDs[i + 1]
+
+        var $theID = $(theID)
+        var $theNextID = $(theNextID)
+
+        var headingPosition = self.getHeadingPosition($theID).top
+        headingPosition = headingPosition - 53 // fix the offset from top of page
+
+        if (theNextID) {
+          var nextHeadingPosition = self.getNextHeadingPosition($theNextID).top
+        }
+
+        var distanceBetweenHeadings = self.getDistanceBetweenHeadings(headingPosition, nextHeadingPosition)
+        if (distanceBetweenHeadings) {
+          var isPastHeading = (windowVerticalPosition >= headingPosition && windowVerticalPosition < (headingPosition + distanceBetweenHeadings))
+        }
+        // when distanceBetweenHeadings is false (as there isn't a next heading)
+        else {
+          var isPastHeading = (windowVerticalPosition >= headingPosition && windowVerticalPosition < footerPosition.top)
+        }
+
+        if (isPastHeading) {
+          self.setActiveItem(theID)
+        }
+
+      })
+
+    }
+
+    self.setActiveItem = function (theID) {
+      self.$anchors.removeClass('active')
+      self.$anchors.filter("[href='" + theID + "']").addClass('active')
+    }
+
+    self.removeActiveItem = function () {
+      self.$anchors.removeClass('active')
+    }
+  }
+})(window.GOVUK.Modules, window)
--- a/app/assets/javascripts/modules/highlight-active-section-heading.js
+++ b/app/assets/javascripts/modules/highlight-active-section-heading.js
@@ -108,7 +108,13 @@
         var $theID = $(theID)
         var $theNextID = $(theNextID)
 
-        var headingPosition = self.getHeadingPosition($theID).top
+        var headingPosition = self.getHeadingPosition($theID)
+
+        if (!headingPosition) {
+          return
+        }
+
+        headingPosition = headingPosition.top
         headingPosition = headingPosition - 53 // fix the offset from top of page
 
         if (theNextID) {
--- a/spec/javascripts/highlight-active-section-heading-spec.js
+++ b/spec/javascripts/highlight-active-section-heading-spec.js
@@ -1,0 +1,95 @@
+/* eslint-env jasmine */
+/* eslint-disable no-multi-str */
+
+describe('A highlight active section heading module', function () {
+  'use strict'
+
+  var module
+  var $element
+
+  beforeEach(function () {
+    module = new GOVUK.Modules.HighlightActiveSectionHeading()
+
+    $element = $('<div class="grid-row" data-module="highlight-active-section-heading">\
+      <div class="column-third">\
+        <div class="page-contents js-page-contents js-stick-at-top-when-scrolling">\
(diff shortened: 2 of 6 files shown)


spec/javascripts/highlight-active-section-heading-spec.js
```diff
@@ -0,0 +1,111 @@
...
+          return {
+            scrollTop: 300
+          }
+        }
+        module.start($element)
+        var $anchor = $element.find('.js-page-contents a')
+        expect($anchor.hasClass('active')).toBe(true)
+      })
+
+      it('highlights a nav item, when the setActiveItem function is passed a href', function () {
```
thehenster: Is this test covered by the test above this one?

spec/javascripts/highlight-active-section-heading-spec.js
```diff
@@ -0,0 +1,111 @@
...
+      })
+
+      it('highlights a nav item, when the page is scrolled', function () {
+        module.getWindowPositions = function () {
+          return {
+            scrollTop: 300
+          }
+        }
+        module.start($element)
+        var $anchor = $element.find('.js-page-contents a')
```
thehenster: It'd be good if we were more explicit about which link has the 'active' class, rather than assert that one of the links is active.

spec/javascripts/highlight-active-section-heading-spec.js
```diff
@@ -0,0 +1,111 @@
...
+              <li>who your likely users are</li>\
+              <li>what they’re trying to do</li>\
+              <li>how they’re trying to do it now</li>\
+            </ul>\
+            <p>Unlike market research, you need to work out how your users think and behave rather than find out what they like or think they want.</p>\
+            <p>If you understand your users’ circumstances, influences and expectations, it will help you design a service that’s easy to use and valued by the people who need it.</p>\
+          </div>\
+        </div>\
+      </div>\
+    </div>')
```
thehenster: I'm wondering how much of this fixture is useful. We're faking the positions of the window scroll and elements.

Maybe a more minimal fixture would:

a) make it easier to read what markup is required
b) would be just as valid a test

spec/javascripts/highlight-active-section-heading-spec.js
```diff
@@ -0,0 +1,111 @@
+/* eslint-env jasmine */
+/* eslint-disable no-multi-str */
+
+describe('A highlight active section heading module', function () {
+  'use strict'
+
+  var GOVUK = window.GOVUK
+  var $ = window.$
```
thehenster: What does this do? Was `$` out of scope without it?

gemmaleigh: I've removed this, it was copied from another example.

spec/javascripts/highlight-active-section-heading-spec.js
gemmaleigh: I've simplified the fixture to make it easier to read.

spec/javascripts/highlight-active-section-heading-spec.js
gemmaleigh: I've added a check here for the third item, as this has the active class.

spec/javascripts/highlight-active-section-heading-spec.js
gemmaleigh: Yes, it's a duplicate - I have removed it.

spec/javascripts/highlight-active-section-heading-spec.js
```diff
@@ -0,0 +1,92 @@
...
+
+      it('highlights a nav item, when the page is scrolled', function () {
+        module.getWindowPositions = function () {
+          return {
+            scrollTop: 300
+          }
+        }
+        var testHref = '#section-3'
+        module.start($element)
+        isLinkHighlighted(testHref)
```
thehenster: We could probably pass the string literal directly rather than assign it to a variable here.

spec/javascripts/highlight-active-section-heading-spec.js
```diff
@@ -0,0 +1,92 @@
...
+            <h2 id="section-1">Section 1</h2>\
+            <p>Section 1 text</p>\
+            <h2 id="section-2">Section 2</h2>\
+            <p>Section 2 text</p>\
+            <h2 id="section-3">Section 3</h2>\
+            <p>Section 3 text</p>\
+          </div>\
+        </div>\
+      </div>\
+    </div>')
```
thehenster: Nice :+1:

spec/javascripts/highlight-active-section-heading-spec.js
```diff
@@ -0,0 +1,92 @@
...
+      it('has no highlighted nav items, when the page loads', function () {
+        module.getWindowPositions = function () {
+          return {
+            scrollTop: 0
+          }
+        }
+        module.start($element)
+        var $anchors = $element.find('.js-page-contents a')
+        expect($anchors.hasClass('active')).toBe(false)
+      })
```
thehenster: Just a style thing for fun.. It's quite popular to break tests into paragraphs - setup, exercise and verification paragraphs. It just means putting a line break in between the paragraphs.

eg.

```
it('has no highlighted nav items, when the page loads', function () {
  setWindowScroll(0)

  module.start($element)

  var $anchors = $element.find('.js-page-contents a')
  expect($anchors.hasClass('active')).toBe(false)
})
```

I also think it'd be cool to hide the complexity of stubbing the window and element position things in a helper function. I had a go in the above example.

None of this essential at all.. Just a suggestion for how I would find it more readable.

app/assets/javascripts/modules/highlight-active-section-heading.js
```diff
@@ -0,0 +1,136 @@
...
+    self.getElementOffset = function ($el) {
+      return $el.offset()
+    }
+
+    self.start = function ($el) {
+      $window.resize(self.hasResized)
+      $window.scroll(self.hasScrolled)
+
+      setInterval(self.checkResize, _interval)
+      setInterval(self.checkScroll, _interval)
```
36degrees: Is there a reason we're polling at intervals rather than binding directly to the scroll events? If it's to throttle/debounce the scroll event, have we ruled out using e.g. https://github.com/cowboy/jquery-throttle-debounce?

Either way, it'd be good to document why we're doing this as it seems a bit odd to me and difficult to follow…

app/assets/javascripts/modules/highlight-active-section-heading.js
```diff
@@ -0,0 +1,136 @@
...
+      }
+    }
+
+    self.getElementOffset = function ($el) {
+      return $el.offset()
+    }
+
+    self.start = function ($el) {
+      $window.resize(self.hasResized)
+      $window.scroll(self.hasScrolled)
```
36degrees: This is totally personal preference, but I've moved to using e.g. `.on('scroll', …)` as I think it makes it a million times clearer that you're setting up a new binding, vs telling the window to resize or scroll itself.

app/assets/javascripts/modules/highlight-active-section-heading.js
```diff
@@ -0,0 +1,136 @@
...
+
+      setInterval(self.checkResize, _interval)
+      setInterval(self.checkScroll, _interval)
+
+      self.$anchors = $el.find('.js-page-contents a')
+      self.totalAnchors = self.$anchors.length
+      self.getAnchors(self.totalAnchors)
```
36degrees: getAnchors is returning anchorIDs, but the result just gets thrown away? I'd suggest either removing the `return` from `getAnchors` to make it clear it's manipulating a variable outside of its own scope, or assigning the result to anchorIDs here.

app/assets/javascripts/modules/highlight-active-section-heading.js
```diff
@@ -0,0 +1,136 @@
...
+
+      self.checkResize()
+      self.checkScroll(self.totalAnchors)
```
36degrees: checkScroll isn't expecting any arguments?

app/assets/javascripts/modules/highlight-active-section-heading.js
```diff
@@ -0,0 +1,136 @@
...
+        } else {
+          self.updateActiveNavItem()
+        }
+      }
+    }
+
+    self.getAnchors = function (totalAnchors) {
+      for (var i = 0; i < self.totalAnchors; i++) {
+        var anchor = self.$anchors[i]
+        var anchorID = $(anchor).attr('href')
```
36degrees: If $anchors is an array of jQuery objects, do you need to call jQuery again? `anchor.attr`should work? If so, worth naming `anchor` `$anchor` instead?

app/assets/javascripts/modules/highlight-active-section-heading.js
```diff
@@ -0,0 +1,136 @@
...
+
+    self.getDistanceBetweenHeadings = function (headingPosition, nextHeadingPosition) {
+      var distanceBetweenHeadings = (nextHeadingPosition - headingPosition)
+      return distanceBetweenHeadings
+    }
+
+    self.updateActiveNavItem = function () {
+      var windowVerticalPosition = self.getWindowPositions().scrollTop
+
+      for (var i = 0; i < self.totalAnchors; i++) {
```
36degrees: Any reason we're opting for this vs jQuery's .each, as we have jQuery loaded anyway?

36degrees:
The commits since my last review all look good. I think my first two points are still outstanding though?

36degrees:
There are a few really picky edge cases:

- If you scroll really quickly to e.g. the top of the document it doesn't quite keep up and you can end up with the wrong thing highlighted.
- If you have a very short section at the end of your document it'll never become active as the previous heading doesn't leave the screen. This is especially odd if you click the heading to scroll down as the previous link becomes active instead.

However I think those really are edge cases and I'd be happy for this to go out as is.

app/assets/javascripts/modules/highlight-active-section-heading.js
gemmaleigh: I was following the example over here [sticky-element-container.js](https://github.com/alphagov/static/blob/master/app/assets/javascripts/modules/sticky-element-container.js#L45), as it works in a similar way. If there's a better way to do this I am happy to amend.

gemmaleigh:
This can't be merged just yet, the govuk_frontend_toolkit's `stick-at-top-when-scrolling.js` needs updating to resize the shim it inserts when the `.content-fixed` class is added.

The bug spotted by @36degrees, means that when the screen is resized, the shim isn't, so the page contents contents is wider than it should be (as the shim prevents the grid-column from resizing the content), leading to the main content overlapping the page contents, at widths wider than 768px.

gemmaleigh:
Also the stick-at-top-when-scrolling.js must be removed from this repo (it's in app/assets/javascripts`) and application.js updated to refer to the file from the frontend toolkit.

gemmaleigh:
I'm now checking for the width of the parent element (in this case `.column-third`) and using this to set the width of the element (`.page-contents`) and the shim `.shim`. 

This can't be added to the front end toolkit as-is (as there's no guarantee a sticky element will exist within a grid column).

Will have to have a think about how best to add this to our app. It may well be better to use our own version of stick-at-top-when-scrolling in this case.

36degrees:
The only thing I've thought of is to modify the Javsacript in front end toolkit to look for a data attribute (or extra class, or similar) on the 'stuck thing' and change its behaviour based on the presence of that attribute.

But I don't object to just adding a modified version to our app if that makes more sense for now.

Either way, it'd be good to see this merged!

gemmaleigh:
I've also adjusted the margins for the govspeak content, so when an item in the left hand navigation is selected, the "Page contents" title and the section heading line up.

Before screenshot:
before

After screenshot:
who-the-community-after

36degrees:
Looks good to me!

gemmaleigh:
It would be great to get this merged! @36degrees could you please deploy it to integration?

PR #62, opened 2016-10-06
Service manual header improvements

36degrees:
Remove the government navigation and add a 'Service Manual' title to ground user's context and help them understand where they are.

**~~Needs updating once a new version of Slimmer has been released with [the related change](https://github.com/alphagov/slimmer/pull/167).~~**

oct-06-2016 09-39-44

# Mobile (< 380px)
bs_realios_mobile_iphone 6-8 0

# Mobile (>= 380px)
bs_realios_mobile_iphone 6s plus-9 0

# Tablet
bs_realios_tablet_ipad air-7 0

# IE8
bs_winxp_ie_8 0


app/assets/stylesheets/modules/_header-tweaks.scss
```diff
@@ -0,0 +1,39 @@
...
+  // search toggle is visible) it needs some padding
+  @media screen and (max-width: 379px) {
+    margin: 5px 15px 0;
+  }
+
+  // When it appears next to the crown, align the heading with the baseline of
+  // the GOV.UK logo...
+  // ...at 16px
+  @media screen and (min-width: 380px) {
+    margin-top: 14px;
```
36degrees: I really don't like these magic numbers, but couldn't find a better way to align the bottom of the text with the baseline of the GOV.UK logo. Suggestions on a postcard.

36degrees:
@robinwhittleton had a query about whether there was something to clear up now that we're no longer using the government navigation.

After a bit of research, the [government_navigation component](http://govuk-component-guide.herokuapp.com/components/government_navigation) comes from [static](https://github.com/alphagov/static/blob/master/app/views/govuk_component/government_navigation.raw.html.erb) and we're able to render it because of [some magic we include from Slimmer](https://github.com/alphagov/slimmer/blob/master/lib/slimmer/shared_templates.rb). We use other govuk components, namely breadcrumbs, the phase labels and analytics meta tags:

```
$ ag govuk_component
app/helpers/phase_label_helper.rb
6:      render partial: "govuk_component/#{presented_object.phase}_label", locals: locals

app/views/content_items/service_manual_guide.html.erb
9:<%= render 'govuk_component/breadcrumbs', breadcrumbs: @content_item.breadcrumbs %>
79:      <%= render 'govuk_component/govspeak', content: @content_item.body.html_safe %>

app/views/content_items/service_manual_service_standard.html.erb
9:<%= render 'govuk_component/breadcrumbs', breadcrumbs: @content_item.breadcrumbs %>

app/views/content_items/service_manual_topic.html.erb
9:<%= render 'govuk_component/breadcrumbs', breadcrumbs: @content_item.breadcrumbs %>

app/views/layouts/application.html.erb
11:  <%= render partial: 'govuk_component/analytics_meta_tags', locals: { content_item: @content_item.content_item } %>

test/integration/phase_label_test.rb
43:    assert page.has_no_css?("[data-template='govuk_component-alpha_label']")
44:    assert page.has_no_css?("[data-template='govuk_component-beta_label']")
48:    within "[data-template='govuk_component-#{phase}_label']" do
```

So, in conclusion, I don't think there's anything to tidy up because we're still using other GOV.UK components and even if we weren't we're still using Slimmer anyway.

app/assets/stylesheets/modules/_header-tweaks.scss
thehenster: My only suggestion would be to assign it to a variable with an comment like name. An extreme example would be `$difference-in-height-between-title-and-the-logo = 14px;` or something, and remove the comment. That _might_ be better.. but only just if so. I'd probably leave it.

app/views/layouts/application.html.erb
```diff
@@ -14,6 +14,9 @@
   <% end %>
 </head>
 <body>
+  <div class="slimmer-inside-header">
+      <h1 class="header-title">Service Manual</h1>
```
thehenster: We're trying to follow the BEM CSS pattern in this repo. `header-title` might be the block name here, with an element name of `name`. So the more BEM like CSS class could be `header-title__name`. Or `service-manual-header__title`? It's kind a case of what _feels_ right here..

Another convention we're following is for the filename to follow the block name. So for a block called 'header-title' the filename would be `_header-title.scss`.

Perhaps the the `#global-header` override stuff could be broken into it's own file which wouldn't follow the BEM pattern because it relates to markup outside of this app.

36degrees: I'm going to leave this as is, because I can't see the header title staying if and when we're able to get rid of the rest of the header 'hacks'.

Also, IMO the use of `.header-title__name` (or similar) implies that there is a block called `.header-title` – which there isn't.

thehenster:
Go for it :+1:

PR #63, opened 2016-10-06
Add homepage links to the standard and communities

36degrees:


test/integration/homepage_test.rb
```diff
@@ -26,4 +26,28 @@ class HomepageTest < ActionDispatch::IntegrationTest
...
+    setup_and_visit_example('service_manual_homepage', 'service_manual_homepage')
+
+    assert page.has_content? <<-TEXT
+      You can view the communities of practice to find more learning
+      resources, see who has written the guidance in the manual and connect
+      with digital people like you from across government.
+    TEXT
+
+    assert page.has_link? 'communities of practice', href: '/service-manual/communities'
+  end
```
thehenster: For me personally, I'd be tempted to remove the assertions on the description from these tests, and maybe just delete the tests altogether.

thehenster: Actually.. is the plan that this hardcoded content will come from the content store later? If so - good.

36degrees: I think so. That's not to say that the text won't change before then. Do you think the descriptions should be removed? My only concern is that the Digital Service Standard link already exists on the page, so the test for that would pass even if this addition didn't exist. Unless I use a `within`, or we just sack these tests off all together.

PR #66, opened 2016-10-21
Don’t use HTML5 browser ‘required’ validation

36degrees:
The HTML5 validation is provided by the browser and we are not able to style the error messages. This means that the validation experience is inconsistent with other parts of GOV.UK and with any more advanced validation that might happen server side.

There is already server side validation for the description being required, so the simplest fix is just to remove the required attribute and fall back to that.

https://trello.com/c/sIU9cdr1/464-consistent-validation-experience-with-improve-this-page-form

### DIFF
--- a/app/views/shared/_improve_this_page.html.erb
+++ b/app/views/shared/_improve_this_page.html.erb
@@ -21,7 +21,7 @@
           <div class="improve-this-page__section">
             <label>
               <span class="improve-this-page__label">How should we improve this page?</span><span class="visually-hidden">required</span>
-              <textarea name="description" rows="5" required></textarea>
+              <textarea name="description" rows="5"></textarea>
             </label>
           </div>
           <p>Include your name and email address if you'd like us to get back to you.</p>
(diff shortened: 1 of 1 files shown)


gemmaleigh:
I think we should keep the required attribute as a pointer for assistive tech and instead use the novalidate attribute on the form.

36degrees:
That being the case, guessing we should also get rid of `<span class="visually-hidden">required</span>`?

I'm guessing that'd cause duplication of 'required' by assistive tech otherwise?

gemmaleigh:
Ah, we already use both.

I'm not sure, I'll test this.

gemmaleigh:
It does read out required twice. Yes, best to remove the <span class="visually-hidden">required</span>, so that it is only read out once.

36degrees:
Thanks, @gemmaleigh! Assume you're happy for this to be merged now?

gemmaleigh:
Hi @36degrees! 

I asked @cfq about it this morning, we both think that it might be useful to have the `required` attribute as well as aria-required="true" as a _belt and braces_ approach for AT in older browsers, but we'd like to get some evidence to support this assumption.

This seems like a fair compromise for now, until we have more information.

PR #68, opened 2016-10-25
Fix the improve-this-page prompt for smaller screens

gemmaleigh:
Before:

before

After:

fix

This PR fixes the prompt for feedback on the page, it's currently broken for smaller screens.
Clearing the parent wrapper (.js-prompt), fixes this.

Add a .clearfix helper class, use this to extend the contain-floats placeholder from the
govuk_frontend_toolkit.

### DIFF
--- a/app/assets/stylesheets/mixins/_helpers.scss
+++ b/app/assets/stylesheets/mixins/_helpers.scss
@@ -3,3 +3,8 @@
 @function em($px, $base: 19) {
   @return ($px / $base) + em;
 }
+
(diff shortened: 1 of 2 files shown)


thehenster:
Thanks @gemmaleigh

Should we be extending the `.improve-this-page` class instead?

gemmaleigh:
I don't like using @extend in this way, it creates associations between classes which shouldn't be there.

https://www.sitepoint.com/avoid-sass-extend/

I'm only extending the placeholder from the frontend toolkit as this is a quick-fix with the code we have at the moment. 

Ideally, this would be a mixin and we could include it in the `.improve-this-page` module.

PR #72, opened 2016-11-15
SurveyMonkey custom variable

thehenster:
The phase label has a link to fill out a surveymonkey survey. If we add a custom variable with the request path then it'll be possible to see which page the user came from in the surveymonkey analytics.

https://trello.com/c/d8dDWj2i/492-update-survey-monkey-code-on-service-manual-to-return-url-slug-as-custom-dimension


NeilvB:
Looks good. This matches what we use on static for the site-wide surveys.

PR #75, opened 2016-11-22
Add anchors to section headings, automatically expand linked to sections

36degrees:
Allow linking directly to a topic section by adding anchors to all section headings, and modify the accordion Javascript to automatically expand any linked-to section.

The primary use for this is when redirecting from pages in the old manual, which best correspond to a topic section. We don't just want to redirect users to the topic as it may not be obvious why it relates to what they were originally trying to view.

### DIFF
--- a/app/assets/javascripts/modules/accordion-with-descriptions.js
+++ b/app/assets/javascripts/modules/accordion-with-descriptions.js
@@ -25,6 +25,7 @@
 
       closeOpenSections();
       checkSessionStorage();
+      openLinkedSection();
 
       bindToggleForSubsections();
       bindToggleOpenCloseAllButton();
@@ -36,7 +37,6 @@
       function replaceSpacesWithUnderscores(str) {
         return str.replace(/\s+/g,"_");
       }
-
 
       function serviceManualTopicPrefix() {
         var topic = getserviceManualTopic();
@@ -79,6 +79,27 @@
       function closeOpenSections() {
         var $subsectionContent = $element.find('.subsection__content');
         closeSection($subsectionContent);
+      }
+
+      function openLinkedSection() {
+        var anchor = getActiveAnchor(),
+            section;
+
+        if (!anchor.length) {
+          return;
+        }
+
+        section = $element.find(anchor)
+          .parents('.subsection')
+          .find('.subsection__content');
+
+        if (section.length) {
+          openSection(section);
+        }
+      }
+
+      function getActiveAnchor() {
+        return GOVUK.getCurrentLocation().hash;
       }
 
       function checkSessionStorage() {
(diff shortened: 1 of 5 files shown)


app/assets/javascripts/modules/accordion-with-descriptions.js
```diff
@@ -81,6 +81,28 @@
...
+      function openLinkedSection() {
+        var anchor = getActiveAnchor(),
+            target,
+            section;
+
+        if (!anchor.length) {
+          return;
+        }
+
+        section = $element.find('#' + anchor)
```
mcmillan: You could probably change this to `anchor` and negate the `substr` in the `getActiveAnchor` method below. Might be worth checking `.hash` always includes a # in all browsers, mind.

spec/javascripts/accordion-with-descriptions-spec.js
```diff
@@ -2,215 +2,242 @@ describe('An accordion with descriptions module', function () {
...
+      $element = $(html);
+      spyOn(GOVUK, 'getCurrentLocation').and.returnValue({
+        hash: '#topic-section-one'
+      });
+      accordion.start($element);
     });
 
-    it("has its state removed in session storage", function () {
-      var GOVUKServiceManualTopic = "GOVUK_service_manual_agile_delivery";
+    it("opens the linked to topic section", function () {
```
thehenster: Opening the section if it's been linked to is a neat idea! It's not currently working on integration :( I get scrolled to the correct title but it doesn't open.

spec/javascripts/accordion-with-descriptions-spec.js
```diff
@@ -2,215 +2,242 @@ describe('An accordion with descriptions module', function () {
...
+        </div>\
+      </div>\
+    </div>';
 
   afterEach(function() {
     $(document).off();
   });
 
-  // Setup
+  describe('when no sub section is linked to', function () {
```
36degrees: It's to allow for two different `beforeEach` behaviours. I need to mock out `getCurrentLocation` before `start` is called, but only in the specific tests. If you have a 'global' multiple `beforeEach` and a nested one they _both_ get called and then you have two lots of setup (the first is just never used) and that feels weird and is hard to debug…

One alternative I guess would be to get rid of the `beforeEach` setup and do it in every single test individually?

Open to suggestions.

thehenster: Ah, OK. You know my preference is always to have the least amount of stuff happening outside of the example as possible. Just moving that `accordion.start($element);` inside each test would do it.

spec/javascripts/accordion-with-descriptions-spec.js
thehenster: You're right - that was it. It works.

When you click on a section to open it, the state is stored in the cookie so that the section is still open when you return to it. This isn't happening with these anchor links. It's probably not a super important feature and the user would be very unlikely to notice.. but it's slightly inconsistent. It might just be a case of calling the "remember the state" function.

thehenster: Yeah, using the back button is probably the most likely way that the user would benefit from the "remembering". That feels OK to me.

PR #76, opened 2016-11-23
Hide "improve this page" bar and form when printing

mcmillan:
At the moment, when printing a page from the service manual that has the
"Is this page useful?" bar at the bottom, it prints the feedback form.

As a printed out feedback form is by and large not particularly useful
to our users, this adds a "print" stylesheet which hides the
`improve-this-page` container.


36degrees:
In an ideal world it'd be good to have the print behaviour encapsulated within the improve-this-page component.

We could use media queries instead, but the hiding behaviour wouldn't apply in IE8. I think that's probably OK, given printing from IE8 would still be functional, it'd just include the form?

I've 'solved' this in the tech docs by using [mixins](https://github.com/alphagov/tech-docs-template/blob/b5a5f394c5db1c7d909e17a9f8386481447d21d7/source/stylesheets/govuk_frontend_toolkit/_conditionals.scss#L29-L39) to conditionally compile [screen and print behaviour](https://github.com/alphagov/tech-docs-template/blob/master/source/stylesheets/modules/_header.scss#L70-L93) within components to two different files… I'm not 100% that's actually going to stick around and it seems like massive overkill here…

mcmillan:
@36degrees Agreed – this was my original approach.

Unfortunately, the `application` stylesheet is set to `media="screen"` by default, and changing that to `media="all"` results in some rather fruity behaviour, particularly around the beta banner and page updates list:

screen shot 2016-11-23 at 13 52 27

I'm happy to use a similar style of mixins here, but (annoyingly) it's going to involve fixing a bunch of issues in existing components, or wrapping everything that exists in `@include screen`, which feels a little unpleasant. Might be I'm being a complete dunce, so please correct me if so!

robinwhittleton:
IE6+ happily do print media statements I believe, I’d go with that route. It’s media queries that requires IE9+.

PR #77, opened 2016-11-29
Remove uneccesary `describe` block

36degrees:
Move `accordion.start($element)` inside each individual test so that we don’t need to define two different widely-scoped `beforeEach` callbacks.

As requested by @thehenster in #75


thehenster:
:+1: Thanks!

36degrees:
For future reference, the [diff is a lot clearer without whitespace](https://github.com/alphagov/service-manual-frontend/pull/77/files?w=0).

PR #90, opened 2017-01-05
Add bare bones service toolkit page

36degrees:
Adds a service toolkit presenter and view.

I've also taken the opportunity to remove the pointless `service_manual_` prefix that's throughout this application which is the bulk of this diff – worth stepping through the commits one by one.

Relies on alphagov/govuk-content-schemas#473 being merged before this will pass on CI.

https://trello.com/c/GHY41UJi/546-create-service-toolkit-page


36degrees:
I clicked around a guide, topic, the homepage, the service standard whilst using the dummy content store. All seemed to work OK. Suggest we test relatively thoroughly on integration.

mcmillan:
:+1: :shipit:

PR #91, opened 2017-01-09
Header and footer for Service Toolkit

36degrees:
- Override the header title to 'Service Toolkit'
- Remove the search field
- Use the 'old style' feedback form for the Service Toolkit
- Remove the footer links in both the Service Toolkit and the Service Manual

### DIFF
--- a/app/controllers/content_items_controller.rb
+++ b/app/controllers/content_items_controller.rb
@@ -6,17 +6,12 @@ class ContentItemsController < ApplicationController
   rescue_from GdsApi::HTTPForbidden, with: :error_403
 
   def show
-    # The Slimmer middleware is responsible for intercepting the response and
-    # wrapping it with the GOV.UK header and footer. We want to use our own
-    # 'report a problem' pattern, so opt out of the bundled one.
-    set_slimmer_headers(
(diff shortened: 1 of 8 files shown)


app/controllers/content_items_controller.rb
```diff
@@ -3,20 +3,17 @@
...
   rescue_from GdsApi::HTTPForbidden, with: :error_403
 
   def show
-    # The Slimmer middleware is responsible for intercepting the response and
-    # wrapping it with the GOV.UK header and footer. We want to use our own
-    # 'report a problem' pattern, so opt out of the bundled one.
-    set_slimmer_headers(
-      report_a_problem: "false"
-    )
+    slimmer_template :without_footer_links
```
mcmillan: Do we want to disable footers across the service manual, in addition to the service toolkit?

36degrees: Yep, checked this with Dave and he's happy for that to happen. The related commit message was actually written by him and mentions both the toolkit and the manual.

mcmillan: Righto! :+1:

PR #92, opened 2017-01-09
Display hero and links on Service Toolkit

36degrees:
https://trello.com/c/iJi4IFtI/545-main-content-on-page

### DIFF
--- a/app/assets/stylesheets/modules/_collection.scss
+++ b/app/assets/stylesheets/modules/_collection.scss
@@ -1,0 +1,37 @@
+.collection {
+  border-bottom: 1px solid $border-colour;
+
+  padding-bottom: $gutter-half;
+  margin-bottom: $gutter-half;
+
+  @include media(tablet) {
+    padding-bottom: $gutter;
+    margin-bottom: $gutter;
+  }
+
+  &:last-child {
+    margin-bottom: 0;
+    border-bottom: none;
+  }
+
+  &__title {
+    @include bold-36;
+    margin-bottom: 5px;
+    margin-top: 1.25em;
+  }
+
+  &__description {
+    @include core-24;
+    color: $secondary-text-colour;
+    margin-top: 0;
+  }
+
+  &__link {
+    @include bold-24;
+    margin: 0.8333333333em 0;
+  }
+
+  &__link-description {
+    @include core-19;
+  }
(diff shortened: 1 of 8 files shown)


app/presenters/content_item_presenter.rb
```diff
@@ -8,13 +8,20 @@ def initialize(content_item)
...
     @locale = content_item["locale"] || "en"
     @phase = content_item["phase"]
-    @links = content_item["links"] || {}
   end
 
   def available_translations
     sorted_locales(links["available_translations"])
   end
 
+  def links
```
mcmillan: Is there any reason we need a method to do this? It feels like it might be easier to put...

```
@links = content_item["links"] || {}
@details ||= content_item["details"] || {}
```

...in the `initialize` method of this class, and include `attr_reader :links, :details` if they need to be accessed via methods.

mcmillan: Sorry. Right. I see, they've come from elsewhere. Feel free to ignore, but might be a nice refactor.

app/assets/stylesheets/modules/_collection.scss
```diff
@@ -0,0 +1,37 @@
...
+
+  &__description {
+    @include core-24;
+    color: $secondary-text-colour;
+    margin-top: 0;
+  }
+
+  &__link {
+    @include bold-24;
+    margin: 0.8333333333em 0;
```
mcmillan: How are we arriving at this value? It might be worth adding a comment to explain what it means.

app/presenters/content_item_presenter.rb
36degrees: No, I just moved them from the [guide presenter](https://github.com/alphagov/service-manual-frontend/pull/92/files#diff-fe302837b4e13df9599084e6ef3366e3) to make them available to the toolkit presenter.

36degrees: Happy to make that change, but let's tidy up once this stack of branches is merged.

app/assets/stylesheets/modules/_collection.scss
36degrees: 20px (20 / 24 = 0.8333333333). But yeah, should be clearer.

PR #99, opened 2017-01-18
Fix vertical alignment

alextea:
Removing the top margin for the first service standard point vertical aligns it with the notifications sidebar

**Before:**
screen shot 2017-01-18 at 11 52 48

**After:**
screen shot 2017-01-18 at 16 50 38


36degrees:
Have asked @dashouse to take a look as it's more a design change than a code change. Three people reviewing a three line code change. Basically we can take responsibility for a line each 🎉

PR #100, opened 2017-01-24
Add schema test support to Jenkinsfile

mcmillan:
This adds support for running schema tests in govuk-content-schemas.

For more information, see [here](https://github.gds/gds/opsmanual/pull/862/files?short_path=db37eb8#diff-db37eb8a3f2b8b7408aa69b7a0435c61).

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -20,9 +20,29 @@ node {
       paramsToUseForLimit: REPOSITORY,
       throttleEnabled: true,
       throttleOption: 'category'],
+    [$class: 'ParametersDefinitionProperty',
+      parameterDefinitions: [
+        [$class: 'BooleanParameterDefinition',
+          name: 'IS_SCHEMA_TEST',
+          defaultValue: false,
+          description: 'Identifies whether this build is being triggered to test a change to the content schemas'],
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -31,7 +51,7 @@ node {
 
       govuk.setEnvar("RAILS_ENV", "test")
       govuk.bundleApp()
-      
+
       govuk.contentSchemaDependency()
```
suzannehamilton: You should pass `env.SCHEMA_BRANCH` as a parameter to this function so that the correct branch of the schemas is checked out. Otherwise it defaults to `deployed-to-production`, which is correct for branch builds of _this_ project, but incorrect for the schema tests.

I realise I forgot to add this to the service manual instructions, and I also missed it in some of the Jenkinsfiles that I created! 😱  I'll go back and update them, but in the meantime here's an example of the correct config: https://github.com/alphagov/collections-publisher/blob/master/Jenkinsfile

Otherwise, this all looks good.

mcmillan: @suzannehamilton Ah! Right, that makes sense. Duly updated, both here and in alphagov/service-manual-publisher#379. Thanks for the review!

PR #103, opened 2017-02-02
In page navigation fixes

36degrees:
See individual commits for details

https://trello.com/c/CdY1MHPs/548-fix-left-sticky-nav-jump-bug-s


Gemfile
```diff
@@ -3,7 +3,7 @@ source 'https://rubygems.org'
 ruby File.read(".ruby-version").strip
 
 gem 'airbrake', '4.0'
-gem 'govuk_frontend_toolkit', '5.0.0'
+gem 'govuk_frontend_toolkit', '~> 5.1.0'
```
36degrees: 5.1.0 will allow the latest patch version (5.1.x) to be installed when running `bundle update` whereas ~5.1.1 would prevent updating at all without editing the Gemfile.

gemmaleigh: Cool! Thanks for explaining :)

PR #105, opened 2017-02-14
Fix JavaScript tests of accordion session storage

suzannehamilton:
Spotted while porting the accordion analytics tracking to collections.

The Jasmine tests were manipulating the session state in the test code, which meant that they were not fully testing that the application code.

This fixes the problem by removing all state changes in the tests except for resetting the session storage after each test.

To support this, some jQuery selectors have also been updated from `$('foo')` to `$element.find('foo')`. This doesn't change the behaviour in the browse, but it makes the code more consistent (since `find` was already being used elsewhere) and it means that the selector is correctly applied to the HTML fixture in the tests.


36degrees:
Thanks!

PR #108, opened 2017-02-17
Allow "improve this page" form to be closed

mcmillan:
At the moment, if I'm shown the feedback form in the improve this page
component, there's no way to close it. This is apparently rather
annoying, particularly on smaller viewports, so this adds a "Close" link
allowing users to close it.


36degrees:
Can we add a test for this behaviour to `/spec/javascripts/improve-this-page-spec.js`?

mcmillan:
@36degrees Tested!

PR #117, opened 2017-05-03
Fix IE7 errors

36degrees:


gemmaleigh:
This looks good! Thanks @36degrees.

PR #120, opened 2017-07-17
Move static pages from design-principles

danielroseman:
The designprinciples app needs to be killed as it is running a very old version of Rails and does not do anything dynamic in any case. This PR moves the three sets of pages into this app:

* /transformation (and sub-pages for each exemplar) - these are completely static pages, including all HTML and precompiled assets, under /public. 
* /designprinciples and /performanceframework - these are "semi-static"; the content is hard-coded in ERB templates but they are served via a controller so they get wrapped by Slimmer. 
* /accessiblepdfs - sole page remaining of the old content guidance pages, semi-static as above. All the other links on this page previously redirected to the new guidance, they have been updated to point there directly.

Also includes a rake task to re-register these routes with publishing-api so they point to this app.


36degrees:
The rake tasks for the other pages we publish (e.g. /service-toolkit) are in service manual publisher. It'd be good to be consistent.

I also think we'll run into issues running the rake task - not being able to talk to the publishing api from the frontend boxes?

fofr:
@36degrees There's a precedent for running rake tasks from frontend applications in calculators and licence-finder.

fofr:
This shouldn't block this PR, however:
Having hardcoded HTML content is something we should avoid.

Content should be markdown and follow govspeak conventions, using components and existing templates to render. I'd hoped these pages could be managed with something more like https://gist.github.com/fofr/ee439b6c75df8665e4d905a846919aef (or an existing publishing tool)

danielroseman:
I was reluctant to put the rake task in service-manual-publisher, because these pages aren't in any sense published by that app. Plus, we'd want to still have the publishing-app as "designprinciples" to avoid path reservation issues, which adds to the inconsistency.

However, it turns out that the issue with having the task here is not accessing the publishing-api, but the lack of a bearer token for this app. I could set one up, but that requires a whole load of puppet and hieradata changes, which seems overkill for this.

The third alternative is just to do this manually - there are only two affected content_ids, this could be run from the console on deploy.

tijmenb:
> Plus, we'd want to still have the publishing-app as "designprinciples" to avoid path reservation issues, which adds to the inconsistency.

I'm not sure if you can publish things as another `publishing_app`. In any case it would be good to use the correct publishing app to avoid confusion later.

The path reservation issues can be fixed with this: https://github.com/alphagov/content-tagger/blob/master/lib/tasks/publishing_api/publish_api_prefix.rake#L10

danielroseman:
The point is that it's not clear what is the "correct" publishing_app in this case.

As far as I know publishing-api doesn't do anything to validate the publishing_app against the bearer token. Thanks for reminding me about the override_existing argument though.

tijmenb:
@danielroseman I'm not sure why it wouldn't be `service-manual-frontend`? It's how we do `publishing_app` in all other special route registrations. Once design principles has been removed it should also be removed from https://github.com/alphagov/govuk-content-schemas/blob/master/formats/definitions.json#L166-L194, which would make this thing invalid.

danielroseman:
I've removed the rake task from this app. I've raised https://github.com/alphagov/design-principles/pull/287 to update the task on designprinciples, and https://github.com/alphagov/govuk-content-schemas/pull/640 to remove design-principles from the schemas and add service-manual-frontend to publishing_apps.

tijmenb:
@danielroseman will we add the rake task back to this app once design-principles is retired?

danielroseman:
I hadn't planned to; it wouldn't be able to run without a bearer token for publishing_api.

tijmenb:
Why not add it? Do you want to keep around the repo & deploy scripts just for publishing a content item?

36degrees:
The content designers on our team have four stories to move the design principles, performance framework, transformation and the exemplars to Whitehall which they're hoping to complete this week. That would make at least some of these changes redundant.

We may still need to set up redirects to the new URLs, but I think that should probably be done at a higher level, not within service manual frontend?

Can we hold off on merging this until the end of the week and review where we are with this then?

danielroseman:
@36degrees did that content get moved?

36degrees:
It's 'in progress' - I know it's actively being worked on at the minute. I _believe_ the content has all been moved, it just needs reviewing.

We'll need to set up redirects from the old URLs to the new ones. Is that something that should go into router-data, or is there a better way of doing that?

danielroseman:
router-data is being deprecated, it should be possible to set them up in short-url-manager. (It will still need some developer intervention because the paths are currently owned by design-principles and I don't think short-url-manager yet supports force-create.)

h-lame:
We can close this.  We retired design-principles in early December 2017 when all the content was redirected to something else or created as new pages using a publishing tool.

PR #122, opened 2017-09-01
Fix alignment of feedback form on Service Toolkit

36degrees:
For the service toolkit, we override the wrapper to make it full width to allow for the full width masthead. This means that it doesn’t center the feedback form and its toggle as it does on other pages on GOV.UK.

Sadly this means we need to add more hacks to explicitly center the feedback form and its toggle when this full width override is in place.

## Before

screen shot 2017-09-01 at 15 12 38

## After

screen shot 2017-09-01 at 15 12 29

### DIFF
--- a/app/assets/stylesheets/hacks/_full-width.scss
+++ b/app/assets/stylesheets/hacks/_full-width.scss
@@ -15,7 +15,9 @@
     border-bottom: none;
   }
 
-  .full-page-width-wrapper {
+  .full-page-width-wrapper,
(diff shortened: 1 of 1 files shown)


36degrees:
Yeah, it might be clearer if it was called something like full-width__page-wrapper as it's the thing that behaves the normal #wrapper when used on an otherwise full width page…

It's all a bit hacky because we're fighting against the norms that static / slimmer assume.

Thanks for the review 😄

PR #153, opened 2018-02-27
Rollup dependabot prs in one

h-lame:
This collects 11 dependabot PRs in one PR.  We also updates the app config and boilerplate code as a result of the rails 5.1 upgrade PR, and bump to govuk_app_config 1.3.1 and deal with the consequences of doing that.

Replaces #150, #149, #147, #145, #142, #140, #135, #134, #131, #130, #129, and provides alternative implmentations of #144, #146, #136.


config/environments/production.rb
```diff
@@ -80,11 +79,5 @@
...
 
   # Use default logging formatter so that PID and timestamp are not suppressed.
-  config.log_formatter = ::Logger::Formatter.new
-
-  $real_stdout = $stdout.clone
-  $stdout.reopen($stderr)
-  config.logstasher.enabled = true
-  config.logstasher.logger = Logger.new($real_stdout)
-  config.logstasher.suppress_app_log = true
+  # config.log_formatter = ::Logger::Formatter.new
```
suzannehamilton: Should this be removed altogether?

h-lame: I left it in commented out because it's a line that is there from a vanilla rails install.  My theory is that this makes it easier to see what's a change we've made vs what's a new bit of config when we do future rails upgrades.

suzannehamilton: 👍

PR #364, opened 2018-12-21
Move wraith to the development group

cbaines:
As wraith depends unnecessarily on an older version of the
chromedriver-helper gem, installing this gem causes issues with
chromedriver working correctly.

Moving this to the development group prevents wraith being
unnecessarily installed, which should work around this issue.


cbaines:
This should unblock https://github.com/alphagov/govuk_test/pull/11

cbaines:
Thanks @36degrees :)

PR #427, opened 2019-04-02
Load all scripts from govuk_publishing_components

alex-ju:
Scripts in govuk_publishing_components contain not only component-related scripts, but also polyfills and accessibility shims that should be included in every application.


alex-ju:
Thanks @andysellick! This will fix a JavaScript error in the feedback component for Service Manual pages.

PR #444, opened 2019-04-30
Fix failing test when run in BST

36degrees:
When `travel_to("2015-10-10")` is called from a test environment in British Summer Time then Time.now is reported as `2015-10-10 00:00:00 +0100` but the tests assume it’ll be `2015-10-10 00:00:00 +0000`.

This means that the time reported in '.metadata' is reported as 'about 15 hours ago' which means the test fails.

I assume that this isn’t being seen on CI because CI lives in UTC(?).


36degrees:
Thanks, @cbaines.

PR #543, opened 2019-09-18
Add pull request template

injms:
## What
Adds a pull request template, which includes list of pages to check.

## Why
To add consistency to the helpful information added to a pull request, and tohopefully avoid any changes being made the cause a broken layout to go live.

## Visual changes
None

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,22 @@
+## What
+<!-- Description of the change being made -->
+
+## Why
+<!-- What are the reasons behind this change being made? -->
+
+## Visual Changes
+<!-- If the change results in visual changes, show a before and after -->
+
+<!--
+## Pages to check
+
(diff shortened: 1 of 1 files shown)


vanitabarrett:
Just trying to get review apps / heroku working and then we can update the links 🤞

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,21 @@
...
+<!-- What are the reasons behind this change being made? -->
+
+## Visual Changes
+<!-- If the change results in visual changes, show a before and after -->
+
+## Pages to check
+
+Locally, as currently no preview app.
+
+* http://service-manual-frontend.dev.gov.uk/service-manual/
```
vanitabarrett: Service Manual Frontend now has review apps, so we can link to these? Not sure how feasible it is to update each link with the PR number for each review app though..

injms: I've borrowed from the Publishing Components template, which has the URL commented out and a `[PUT-PR-NUMBER-HERE]` comment in the URL - not automatic, but I think better than leaving it blank.

PR #569, opened 2019-10-29
Remove govuk_frontend_toolkit sass dependency

andysellick:
## What
Remove all references to `govuk_frontend_toolkit` sass and replace with `govuk-frontend`.

## Why
`govuk_frontend_toolkit` is deprecated.

## Visual Changes
Should be no visual changes, if you spot any, something's not right.

## Pages to check

* https://govuk-service-manual-fr-pr-569.herokuapp.com/service-manual/
* https://govuk-service-manual-fr-pr-569.herokuapp.com/service-manual/helping-people-to-use-your-service
* https://govuk-service-manual-fr-pr-569.herokuapp.com/service-manual/design
* https://govuk-service-manual-fr-pr-569.herokuapp.com/service-manual/service-assessments
* https://govuk-service-manual-fr-pr-569.herokuapp.com/service-manual/service-standard
* https://govuk-service-manual-fr-pr-569.herokuapp.com/service-manual/service-standard/point-1-understand-user-needs
* https://govuk-service-manual-fr-pr-569.herokuapp.com/service-manual/communities
* https://govuk-service-manual-fr-pr-569.herokuapp.com/service-manual/communities/accessibility-community


app/assets/stylesheets/mixins/_margins.scss
```diff
@@ -1,15 +1,15 @@
 @mixin responsive-bottom-margin {
   margin-bottom: govuk-spacing(3);
 
-  @include media(tablet) {
+  @include govuk-media-query($from: tablet) {
```
injms: Could these margins be replaced with the [`govuk-responsive-margin()`](https://github.com/alphagov/govuk-frontend/blob/master/src/govuk/helpers/_spacing.scss#L101)?

app/assets/stylesheets/mixins/_helpers.scss
```diff
@@ -6,5 +6,5 @@
 
 // https://github.com/alphagov/govuk_frontend_toolkit/blob/master/stylesheets/_shims.scss#L45
 .clearfix {
```
injms: Could we do a global search-and-replace in the templates instead of extending the `clearfix` class?

app/assets/stylesheets/application.scss
```diff
@@ -3,8 +3,9 @@ $govuk-compatibility-govukfrontendtoolkit: true;
...
 
 @import "govuk_publishing_components/all_components";
-
-@import "common";
+@import "mixins/helpers";
+@import "mixins/margins";
+@import "mixins/print";
 
 // CSS that could be better, mostly overriding Slimmer...
 @import "hacks/full-width";
```
injms: Love a good honestly named style file: 'hacks' 😄

app/assets/stylesheets/mixins/_margins.scss
andysellick: I'm not sure. The existing mixins give a margin bottom/top of 15px on mobile and whatever 7 is above that (40px?). How would the govuk mixin replace that?

app/assets/stylesheets/mixins/_helpers.scss
andysellick: I've just done the search, and there appears to be no instances of this class anywhere in the app! Odd. Guess I'll take it out completely then...

app/assets/stylesheets/mixins/_helpers.scss
```diff
@@ -3,8 +3,3 @@
 @function em($px, $base: 19) {
```
vanitabarrett: It doesn't seem like this is used anywhere - perhaps we can just delete the whole helpers file?

app/assets/stylesheets/mixins/_margins.scss
injms: Ah, true - if you want it exactly the same you can't.  Would a 25px margin on small screens be acceptable instead of a 15px margin? Then `@include govuk-responsive-padding(7)` could simplify things. Or `5` would give a 25px margin on large screens and 15px on small screens.

andysellick: I'm not familiar enough with the service manual to be comfortable making that change. I think it's definitely something to revisit later though.

app/assets/stylesheets/mixins/_helpers.scss
andysellick: Removed, thanks.

vanitabarrett:
Only had a chance to do a really quick visual check on Heroku, but code looks ok 👍

PR #585, opened 2019-11-27
Remove frontend toolkit from service manual frontend

andysellick:
## What
Remove govuk-frontend-toolkit from service-manual-frontend.

This involves only one noticeable change for users - on [guide pages](https://www.gov.uk/service-manual/helping-people-to-use-your-service/making-your-service-accessible-an-introduction) a script from toolkit was being used to simulate `position: sticky`. This has been removed (well, not removed, but not called from toolkit anymore) and replaced with actual `position: sticky`. 

Sticky is supported in all major browsers apart from IE, but the only difference is that the menu won't scroll with the page. Consensus among the FE devs is that it's not worth the effort of a polyfill (also I tried and there are several problems that make the code hacky, so best avoided).

## Why
Goodnight toolkit, well, it's time to go...

## Visual Changes
No visual changes other than the sticky menu.

## Pages to check

* https://govuk-service-manual-fr-pr-585.herokuapp.com/service-manual/
* https://govuk-service-manual-fr-pr-585.herokuapp.com/service-manual/helping-people-to-use-your-service
* https://govuk-service-manual-fr-pr-585.herokuapp.com/service-manual/design
* https://govuk-service-manual-fr-pr-585.herokuapp.com/service-manual/service-assessments
* https://govuk-service-manual-fr-pr-585.herokuapp.com/service-manual/service-standard
* https://govuk-service-manual-fr-pr-585.herokuapp.com/service-manual/service-standard/point-1-understand-user-needs
* https://govuk-service-manual-fr-pr-585.herokuapp.com/service-manual/communities
* https://govuk-service-manual-fr-pr-585.herokuapp.com/service-manual/communities/accessibility-community

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -1,5 +1,4 @@
 $govuk-compatibility-govuktemplate: true;
-$govuk-compatibility-govukfrontendtoolkit: true;
 $govuk-compatibility-govukelements: true;
 
 @import "govuk_publishing_components/all_components";
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -22,4 +22,5 @@ $govuk-compatibility-govukelements: true;
 @import "modules/panel";
 @import "modules/related-content";
 @import "modules/service-standard-point";
+@import "modules/sticky";
 @import "modules/typography";
--- a/app/views/layouts/application.html.erb
+++ b/app/views/layouts/application.html.erb
@@ -6,7 +6,6 @@
   <%= stylesheet_link_tag "print", media: "print" %>
 
   <%= javascript_include_tag "application" %>
-  <%= javascript_include_tag "start-modules" if Rails.env.test? %>
   <%= csrf_meta_tags %>
   <%= render "govuk_publishing_components/components/meta_tags", content_item: @content_item.content_item %>
   <%= yield(:extra_head) %>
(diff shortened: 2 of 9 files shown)


app/assets/stylesheets/application.scss
```diff
@@ -23,3 +22,10 @@ $govuk-compatibility-govukelements: true;
 @import "modules/related-content";
 @import "modules/service-standard-point";
 @import "modules/typography";
+
+.sticky-element {
```
alex-ju: Can we prefix this with `app-` and move it as a partial within what we call in this app `stylesheets/modules`?

app/views/layouts/application.html.erb
```diff
@@ -6,7 +6,6 @@
   <%= stylesheet_link_tag "print", media: "print" %>
 
   <%= javascript_include_tag "application" %>
-  <%= javascript_include_tag "start-modules" if Rails.env.test? %>
```
alex-ju: 💯

app/assets/stylesheets/application.scss
andysellick: Yes! I thought of moving this somewhere after I created the PR, but wasn't sure where a good place would be so decided to wait for suggestions 😁

alex-ju: Ideally, we'd have an ITCSS-based structure of folders, but this way at least we're making it consistent with the current approach and ease the transition – whenever that will happen.

andysellick: Thanks, changed.

PR #616, opened 2020-01-16
Turn off legacy colour palette to update to new colours

injms:
## What
The legacy palette flag (`$govuk-use-legacy-palette`) is set to false to turn on the new colours.

## Why
Part of the process to make GOV.UK more accessible - card is https://trello.com/c/Dd49EiEv/122-switch-to-new-colours-in-service-manual-frontend

## Visual Changes
Oh so many. 

Any yellow and blue should change from:
image

to:
image

## Pages to check

* https://govuk-service-manual-fr-pr-616.herokuapp.com/service-manual/
* https://govuk-service-manual-fr-pr-616.herokuapp.com/service-manual/helping-people-to-use-your-service
* https://govuk-service-manual-fr-pr-616.herokuapp.com/service-manual/design
* https://govuk-service-manual-fr-pr-616.herokuapp.com/service-manual/service-assessments
* https://govuk-service-manual-fr-pr-616.herokuapp.com/service-manual/service-standard
* https://govuk-service-manual-fr-pr-616.herokuapp.com/service-manual/service-standard/point-1-understand-user-needs
* https://govuk-service-manual-fr-pr-616.herokuapp.com/service-manual/communities
* https://govuk-service-manual-fr-pr-616.herokuapp.com/service-manual/communities/accessibility-community

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -1,5 +1,6 @@
 $govuk-compatibility-govuktemplate: true;
 $govuk-compatibility-govukelements: true;
+$govuk-use-legacy-palette: false;
 
 @import "govuk_publishing_components/all_components";
 @import "mixins/margins";


maxgds:
The search box on https://govuk-service-manual-fr-pr-616.herokuapp.com/service-manual/ doesn't seem to have the right styles applied when it's in focus - missing the black border which I think should apply here. I tried adding .govuk-input to it in browser but that didn't fix it.

36degrees:
Re the search box, it's roughly the same 'style' as we have in the Design System and we don't use a black border there – if you were not able to perceive the contrast between the yellow and the white, the box still effectively 'grows' so you are able to perceive where the focus is.

We should probably revisit the focus style for components on a dark background at some point, but in the mean time I think following that should be OK…

injms:
Because this uses the [search input component](https://components.publishing.service.gov.uk/component-guide/search/for_use_on_govuk_blue_background), the focus state comes from the gem - [I've raised an issue on the components gem repo](https://github.com/alphagov/govuk_publishing_components/issues/1259) as a reminder to fix this, and to see where (if!) anyone is using the search component on a blue background.

PR #639, opened 2020-02-21
Change Service Manual to not use 'breadcrumb for current page' option

RobBond-GDS:
Updated Service manual breadcrumb to not display the last item without a url. This is as per the updated breadcrumb design where the last item should be current pages parent

For reference:
https://docs.google.com/document/d/1OFG1Ln6VeCR_bc0qmUtRwMJU62Ljw4F3HO6hq2yloXk

Trello card:
https://trello.com/c/VorvrMfy/39-change-it-so-service-manual-doesnt-use-breadcrumb-for-current-page-option

## What
- removed guide item with no url
- remove topic title with no url
- removed topic breadcrumb with no url
- updated tests to remove no url last item

## Why
Consistent breadcrumb functionality to show parent as the last item and not the current documents title

## Visual Changes

service manual breadcrumb

### DIFF
--- a/app/presenters/service_standard_presenter.rb
+++ b/app/presenters/service_standard_presenter.rb
@@ -13,7 +13,6 @@ class ServiceStandardPresenter < ContentItemPresenter
   def breadcrumbs
     [
       { title: "Service manual", url: "/service-manual" },
-      { title: title },
     ]
   end
 
--- a/app/presenters/topic_presenter.rb
+++ b/app/presenters/topic_presenter.rb
@@ -7,7 +7,7 @@ class TopicPresenter < ContentItemPresenter
   end
 
   def breadcrumbs
-    parent_breadcrumbs << topic_breadcrumb
+    parent_breadcrumbs
   end
 
   def groups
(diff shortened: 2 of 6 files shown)


app/presenters/topic_presenter.rb
```diff
@@ -7,7 +7,7 @@ def initialize(content_item)
   end
 
   def breadcrumbs
-    parent_breadcrumbs << topic_breadcrumb
```
vanitabarrett: We can also remove the `topic_breadcrumb` function now

app/presenters/service_standard_presenter.rb
```diff
@@ -13,7 +13,6 @@ def points
   def breadcrumbs
     [
       { title: "Service manual", url: "/service-manual" },
-      { title: title },
```
vanitabarrett: The test associated with this logic seems to still be testing for a breadcrumb without a URL. It looks like it's still passing, so not sure what's going on there. Something to look into?

https://github.com/alphagov/service-manual-frontend/blob/master/test/presenters/service_standard_presenter_test.rb#L57

RobBond-GDS: fixed test

app/presenters/topic_presenter.rb
RobBond-GDS: removed function

PR #728, opened 2020-05-29
Repoint email signups to generic API

benthorner:
https://trello.com/c/KNWv9Dgx/258-deprecate-and-remove-legacy-email-signup-pages

Depends on: https://github.com/alphagov/email-alert-frontend/pull/806

Previously we relied on each topic (and the service standard) having
an associated '/email-signup' content item. The matching criteria are
the topic itself, with a restriction on the document type [1]. The
additional document type restriction turns out to be redundant, since
service manual guides are the only document type linked to service
manual topics (and similarly for the service standard). This repoints
the email signup links to the generic page on Email Alert Frontend,
which supports signups to a given content item.

[1]: https://github.com/alphagov/service-manual-publisher/blob/master/app/presenters/topic_email_alert_signup_presenter.rb#L53

## Before

Screenshot 2020-05-29 at 11 56 02
Screenshot 2020-05-29 at 11 56 15

## After

Screenshot 2020-05-29 at 11 55 48
Screenshot 2020-05-29 at 11 56 32

### DIFF
--- a/app/presenters/service_standard_presenter.rb
+++ b/app/presenters/service_standard_presenter.rb
@@ -17,8 +17,7 @@ class ServiceStandardPresenter < ContentItemPresenter
   end
 
   def email_alert_signup_link
-    signups = content_item["links"].fetch("email_alert_signup", [])
-    signups.first["base_path"] if signups.any?
+    "/email-signup?topic=#{content_item['base_path']}"
   end
 
 private
--- a/app/presenters/topic_presenter.rb
+++ b/app/presenters/topic_presenter.rb
@@ -25,8 +25,7 @@ class TopicPresenter < ContentItemPresenter
   end
 
   def email_alert_signup_link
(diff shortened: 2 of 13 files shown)


vanitabarrett:
I think all these changes mean we can remove the old email-signup page code, e.g: https://github.com/alphagov/service-manual-frontend/blob/master/app/views/shared/_email_signup.html.erb

Unless you were planning on dealing with that in a follow-up PR?

benthorner:
I think we still need to keep it, as we're only changing the URL, and not the rendering. See [here](https://github.com/alphagov/service-manual-frontend/pull/728/files#diff-d06cc4bc5f948a60b4b4d2b2e7fe6033R74), for example.

vanitabarrett:
D'oh sorry, got confused with the work Kelvin was doing. Yep that makes sense 👍

PR #749, opened 2020-07-07
Upgrade application to Rails 6.0

pixeltrix:
As well as upgrading the application the testing stack has been updated to use an up-to-date version of Capybara and switch to using Google Chrome in headless mode instead of PhantomJS whose development status is unclear.


Gemfile
```diff
@@ -2,36 +2,35 @@ source "https://rubygems.org"
...
-gem "rails-i18n", "~> 5.1.3"
-gem "rails_translation_manager", "~> 0.1.0"
-gem "sass-rails", "~> 5.0.6"
-gem "slimmer", "~> 15.0.0"
-gem "uglifier", ">= 1.3.0"
+gem "gds-api-adapters"
+gem "govuk_app_config"
+gem "govuk_publishing_components"
+gem "plek"
+gem "rails", "~> 6.0.3", ">= 6.0.3.2"
```
thomasleese, line 9: Do we still need to specify the version for Rails? It might look strange for it to be the only one that's got a specific version.

pixeltrix: This is what Rails currently generates when you do `rails new` so I was following that as a guide. The security level requirement is there because doing `~> 6.0.3.2` would prevent updating to 6.0.4 but we could just leave it as `~> 6.0.3`, drop the second constraint and let the lock file control the security release level.

kevindew: Hey, I got directed here after asking similar question on https://github.com/alphagov/service-manual-frontend/pull/749#discussion_r454243642

Do you see there being any risks if we didn't specify the version and just used the Gemfile.lock for it? The default Gemfile you get with a Rails install has a lot of different version pinnings in it, so it doesn't feel as inconsistent there as it does here (though I do find it raises a number of questions as to why some things are strict and others not). The project we've set no version pinning as an initial experiment, [email-alert-api](https://github.com/alphagov/email-alert-api/blob/49c59c0ffbcd989935a2f027b82e4c5f668d1a3d/Gemfile), has already a taken on no explicit rails pinning approach so we risk creating inconsistencies.

pixeltrix: Without any pinning there is a possibility that rails could go backwards if another gem was added to the Gemfile that had a constraint on `rails < 6` and then the a developer ran `bundle install` - however, it's unlikely that anyone will be adding non-Rails 6 compatible gems at this point and would almost certainly be picked up in code review if the developer missed the bundler warning about version regression.

thomasleese: I guess this is a potential problem with all of our Gems if we go down the route of not specifying a version in the `Gemfile`.

My personal preference would be to remove the version here and rely on the code review to catch downgrades - or to go back to specifying the version for all of our Gems (which I _don't_ think we should do). I prefer the file being consistent otherwise I suspect people will come across this and try to work out why only the version for Rails has been specified.

pixeltrix: My response to "why is only the Rails version specified?" would be to say that it's the root of the dependency graph and so should be the primary influence on the version of everything in that dependency graph, but like you I'd much rather see no versions so happy to remove and rely on code review.

kevindew: Thanks @pixeltrix that Rails downgrade concern makes sense, I can see why Rails core does the greater than condition to avoid you falling behind a security update.

To add some context our concerns around this are because the pinning in our gemfiles so often ends up so inconsistent  that it becomes essentially meaningless, which means very few people pay attention to it when it is meaningful. It'd be nice to get to a point where everything that had a pin had a comment above it to explain why that was special (for example elasticsearch gem needs pinning to the major version of elasticsearch).

It seems good to roll with Rails being unpinned while we've not got evidence of experiencing a problem from it. If we do see a problem I wonder if it'd make sense to go with a fixed one of say `rails "6.0.3.2"` as per Ruby version to remove the aspects of flexibility a range constraint has, which may also emphasis the Rails root dependency special status

pixeltrix: @kevindew @thomasleese so on upgrading another application I found a good reason for including a target version for the Rails gem - this is what I got after I'd unpinned all the dependencies for Rails 5.2 and then tried to update:

```
% bundle update rails
Fetching gem metadata from https://rubygems.org/..........
Fetching gem metadata from https://rubygems.org/.
Resolving dependencies......
...
Bundler attempted to update rails but its version stayed the same
Bundle updated!
```

I then put in `gem "rails", "6.0.3.2"` like you suggested might be workable and got this:

```
% bundle update rails        
Fetching gem metadata from https://rubygems.org/..........
Fetching gem metadata from https://rubygems.org/.
Resolving dependencies....
Bundler could not find compatible versions for gem "activemodel":
  In Gemfile:
    kaminari-mongoid was resolved to 1.0.1, which depends on
      mongoid was resolved to 6.3.0, which depends on
        activemodel (~> 5.1)

    rails (= 6.0.3.2) was resolved to 6.0.3.2, which depends on
      activemodel (= 6.0.3.2)
```

That information then allowed me to do this:

```
% bundle update rails mongoid
Fetching gem metadata from https://rubygems.org/..........
Fetching gem metadata from https://rubygems.org/.
Resolving dependencies....
...
Bundle updated!
```

Without the version the only option is to run a full `bundle update` which I've found to be unhelpful when trying to track down what dependency change has introduced a bug or failing test.

kevindew: @pixeltrix thanks for explaining the problem, I can see why that could be a bit painful.

So shall we try having Rails in a group at it's top with it's version defined singularly?

I guess for this Gemfile that'd come out like:

```
ruby File.read(".ruby-version").strip

gem "rails", "6.0.3.2"

gem "gds-api-adapters"
gem "govuk_app_config"
gem "govuk_publishing_components"
# ...
```

which does come out looking a bit special, which seems to be how we want to treat Rails. That approach works for me, how about you @thomasleese ? Does that also seem a good practice for you @pixeltrix? or does it have any smells?

pixeltrix: Separating out rails at the top with a specific version is how I prefer it

https://github.com/alphagov/e-petitions/blob/master/Gemfile#L6

kevindew: Super, would you be able to leave a summary of the why in the commit for this repo so we can reference it in future? Hopefully we'll get it put into actual documentation once these approaches become slightly more ingrained as conventions.

thomasleese: This approach makes sense to me, thanks for looking into it! 😄

PR #790, opened 2020-10-15
Lint as part of the default rake task

kevindew:
Trello: https://trello.com/c/5Gob6WUL/147-for-everyone-ensure-that-rubocop-runs-as-part-of-bundle-exec-rake-as-one-of-the-default-tasks

We had expected that all apps were linting with the default rake task
and removed the linting step from the Jenkins library [1]. However it
seems this app was missed.

The rails test task always seem to run last when part of the default
group despite any ordering. I've tried to make this at least appear
explicit by defining jasmine:ci before test, even though this is
different for how GOV.UK normally applies it.

[1]: https://github.com/alphagov/govuk-jenkinslib/pull/74

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -4,3 +4,13 @@
 require File.expand_path("config/application", __dir__)
 
 Rails.application.load_tasks
+
+begin
+  require "rubocop/rake_task"
+  RuboCop::RakeTask.new
+rescue LoadError
+  # Rubocop isn't available in all environments
+end
+
+Rake::Task[:default].clear if Rake::Task.task_defined?(:default)
+task default: %i[rubocop jasmine:ci test]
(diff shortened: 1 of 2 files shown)


ChrisBAshton (approved):
Is it an assumption that we need the `lib/tasks/` directory to exist? I know we have `.keep` sometimes because an application expects a directory to exist, but have we actually checked that?

(Either way, won't need a re-review 👍 )

kevindew:
> Is it an assumption that we need the `lib/tasks/` directory to exist? I know we have `.keep` sometimes because an application expects a directory to exist, but have we actually checked that?
> 
> (Either way, won't need a re-review 👍 )

I thought it seemed a bit weird to not have a lib directory at all. I'm not too sure if it errors or not but I thought it was better for consistency.

PR #849, opened 2021-01-20
Make use of gem components where possible

owenatgov:
## What
Update several instances of custom markup or markup directly from the design system.

## Why
This is part of an ongoing effort in the govuk accessibility team to use components in as many places as possible, ensuring that we can more easily deploy bugfixes and accessibility updates across our apps and avoid blind spots in our code coverage.

No visual changes.

[Card](https://trello.com/c/yB2tjxwt/564-update-service-manual-frontend-to-use-components)

### DIFF
--- a/app/views/content_items/topic.html.erb
+++ b/app/views/content_items/topic.html.erb
@@ -30,23 +30,31 @@
           items: @content_item.accordion_content
         } %>
       <% else %>
-        <% @content_item.groups.each_with_index { |link_group| %>
+        <% @content_item.groups.each_with_index do |link_group| %>
           <% if link_group.name.present? %>
             <div class="subsection__header govuk-!-margin-bottom-3">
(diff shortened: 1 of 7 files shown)


app/views/content_items/topic.html.erb
```diff
@@ -30,44 +30,58 @@
...
           <nav role="navigation" aria-labelledby="related-communities" class="related-communities">
-            <ul class="related-item__list govuk-list">
-              <% @content_item.content_owners.each do |content_owner| %>
-                <li class="related-item__list-item govuk-body-s">
-                  <%= link_to content_owner.title, content_owner.href, class: 'govuk-link' %>
-                </li>
-              <% end %>
-            </ul>
+            <%
+              related_items = []
```
chris-gds, line 76: I'm seeing this pattern on some of the stuff i've been doing as well. Might be a case to extend the gem in some way.

chris-gds (approved):

chris-gds:
🚀

app/views/content_items/topic.html.erb
owenatgov: When you say this pattern what do you mean?

Also I've just realised from looking at this again that this could totally be mapped. Will update.

PR #870, opened 2021-03-02
Add missing tests

1pretz1:
Clean up and adds some missing tests for Service Manual Frontend in preparation for enabling continuous deployment.

Main features of the app are tested, the two Javascript modules are both tested and the ruby test coverage is now near 100%:

```
Coverage report generated for Minitest, Unit Tests to /govuk/service-manual-frontend/coverage. 248 / 249 LOC (99.6%) covered. 
```

Trello:
https://trello.com/c/zu7IYR6C/2385-enable-continuous-deployment-for-service-manual-frontend

### DIFF
--- a/app/presenters/topic_presenter.rb
+++ b/app/presenters/topic_presenter.rb
@@ -1,6 +1,9 @@
 class TopicPresenter < ContentItemPresenter
   ContentOwner = Struct.new(:title, :href)
 
+  attr_reader :visually_collapsed
+  alias_method :visually_collapsed?, :visually_collapsed
+
   def initialize(content_item)
     super
     @visually_collapsed = content_item["details"]["visually_collapsed"]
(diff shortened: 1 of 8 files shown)


app/presenters/topic_presenter.rb
```diff
@@ -33,7 +35,7 @@ def phase
   end
 
   def visually_collapsed?
-    @visually_collapsed
+    visually_collapsed
   end
```
thomasleese, line 39:
```suggestion
alias_method :visually_collapsed?, :visually_collapsed
```

This might be an alternative to having the method, but I don't really mind. It's sad we can't define the variable with the question mark.

thomasleese (approved):

app/presenters/topic_presenter.rb
1pretz1: Ohh, didn't think of using an alias - I think it makes more sense than just a simple method so thanks for the suggestion 👍  

Yeah it is a little, especially when the value is a boolean. I suppose Rails developers normally get round this explicitly by using a method like `present?` which isn't as nice.

PR #886, opened 2021-03-31
Update URL and message in the beta banner

martin-jordan:

## What
This change replaces the current message and link in the beta banner. Instead of referring to a feedback survey, it links to a newly created contact page.

## Why
The survey is no longer used. As the new team behind the Service Manual takes a community-led approach, the newly created contact page lists different ways to get in touch with the team and give feedback on guidance.

## Visual Changes
Only a change of message – no visual change.


app/helpers/phase_label_helper.rb
```diff
@@ -8,7 +8,7 @@ def render_phase_label(presented_object, message)
   end
 
   def survey_url
-    url = "https://www.smartsurvey.co.uk/s/HGLV5/"
+    url = "https://www.gov.uk/service-manual/communities/contact-the-service-manual-and-service-standard-team"
     url += "?c=#{request.path}" if request.path.present?
     url
   end
```
kevindew, line 14:
```suggestion
```

I think we can remove needing this function and just specify the URL in the template.

app/views/shared/_custom_phase_message_homepage.html.erb
```diff
@@ -1 +1 @@
-Complete our quick 5-question survey to <%= link_to "help us improve our content", survey_url, target: "_blank", rel: "noopener noreferrer", class: "govuk-link" %>.
+<%= link_to "Contact the Service Manual team", survey_url, class: "govuk-link" %> if you have feedback, questions or suggestions.
```
kevindew, line 1:
```suggestion
<%= link_to "Contact the Service Manual team", "/service-manual/communities/contact-the-service-manual-and-service-standard-team", class: "govuk-link" %> if you have feedback, questions or suggestions.
```

As per the previous suggestion we can specify the URL here. As the user is already on www.gov.uk convention would be to not specify the absolute path.

kevindew: By the way this template is repeated in: https://github.com/alphagov/service-manual-frontend/blob/5df8cdfcca3332fee44baea1f8dba239523762fd/app/views/shared/_custom_phase_message.html.erb

which would also need editing.

Since these are identical I would suggest deleting the `_custom_phase_message_homepage.html.erb` file and editing https://github.com/alphagov/service-manual-frontend/blob/5df8cdfcca3332fee44baea1f8dba239523762fd/app/views/content_items/homepage.html.erb#L4 to `render 'shared/custom_phase_message', phase: @content_item.phase`

kevindew (commented):
This looks good - nice one on the test fixes. I've put some suggestions in to make this a bit simpler and resolve an issue on pages that aren't the homepage.

martin-jordan (commented):
Reviewed and accepted suggested change to URL reference

kevindew (approved):

PR #950, opened 2021-07-06
Switch to new Design System link styles

danacotoran:
# What
Set `$govuk-new-link-styles: true` in `application.scss` to enable the new link styles from the Design System.

# Why
The new DS link styles are being rolled out across GOVUK.
We've already applied the new styles to the individual components in the component gem. The next step is to enable them on the frontend applications themselves.
This switches `service-manual-frontend` over to the new styles by setting the `govuk-new-link-styles` flag to `true` in the
`application.scss` file. 

The types of pages rendered by this application are primarily content heavy pages. There are no app-level components to speak of, which makes transitioning to the new link styles a fairly straightforward process. I've tested this on the [Example pages rendered by service-manual-frontend](https://docs.publishing.service.gov.uk/apps/service-manual-frontend.html#example-pages-rendered-by-service-manual-frontend) from the dev docs.

------

### DIFF
--- a/app/assets/stylesheets/application.scss
+++ b/app/assets/stylesheets/application.scss
@@ -1,6 +1,7 @@
 $govuk-compatibility-govuktemplate: true;
 $govuk-compatibility-govukelements: true;
 $govuk-use-legacy-palette: false;
+$govuk-new-link-styles: true;
 
 @import 'govuk_publishing_components/govuk_frontend_support';
 @import 'govuk_publishing_components/component_support';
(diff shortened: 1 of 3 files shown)


alex-ju (approved):
We seem to have a few instanced of `app-link--inverse` that could be replaced with `govuk-link--inverse` but not a blocker

danacotoran:
> We seem to have a few instanced of app-link--inverse that could be replaced with govuk-link--inverse but not a blocker

Great catch @alex-ju, have updated 👏🏼

PR #955, opened 2021-07-13
Rewrite README and startup.sh for GOV.UK consistency

kevindew:
Trello: https://trello.com/c/Psf7EzPB/204-move-away-from-startupsh

This makes the README of this application consistent with wider GOV.UK ones with references to GOV.UK Docker and `startup.sh`.

It removes various unusual things from the README that don't follow other GOV.UK precedents such as CSS class naming conventions and Content Store guidance. 

It also removes Wraith as a visual regression tool, on the expectation that this no longer works and not been used in many years. It also fixes some problems with the homepage rendering for Service Manual.

Further information is in the commits.


README.md
```diff
@@ -1,91 +1,26 @@
 # Service Manual Frontend
 
-Service Manual Frontend is a public-facing app to display the service manual formats on GOV.UK.
+A public-facing GOV.UK application used to display the [service manual](https://www.gov.uk/service-manual) using content is fetched from the [Content Store](https://github.com/alphagov/content-store).
```
benthorner, line 3:
```suggestion
A public-facing GOV.UK application used to display the [service manual](https://www.gov.uk/service-manual) with content fetched from the [Content Store](https://github.com/alphagov/content-store).
```

app/controllers/content_items_controller.rb
```diff
@@ -31,24 +31,23 @@ def load_content_item
   end
 
   def present(content_item)
-    class_name = content_item["document_type"].sub(/^service_manual_/, "").classify
-    presenter_name = "#{class_name}Presenter"
-    presenter_class = Object.const_get(presenter_name)
-    presenter_class.new(content_item)
-  rescue NameError
     if content_item["base_path"] == "/"
```
benthorner, line 34: This took me a little while to grasp. Might be worth a comment to say explicitly that without this check we'll fetch the GOV.UK homepage content item and try to render it with the HomepagePresenter.

benthorner: A more robust check would ensure `service_manual_` was present before substituting it, but I guess we can rely on Router to protect the app in production. This is a net improvement and should be enough to avoid surprises in local development.

benthorner (approved):
Nice - the homepage problem was very unexpected!

Just a couple of more minor comments - this is already good to go 👍.

README.md
kevindew: Oh my, how did I miss that 🙈  - thanks.

app/controllers/content_items_controller.rb
kevindew: That's a good idea, I've applied that suggestion.

benthorner: Nice 💯

PR #956, opened 2021-07-13
Remove unnecessary Dockerfile

kevindew:
Trello: https://trello.com/c/Psf7EzPB/204-move-away-from-startupsh

This application is not part of the publishing-e2e-test suite so there
is nothing that actually makes use of this Dockerfile. It is also
incomplete as it lacks the building of assets.

To complete this work the Docker image will need to be removed from
Dockerhub otherwise deployments will be flagged as unstable (due to
Dockerhub no longer being up-to-date). This should have no practical
consequence as nothing should be using these Dockerhub images.


benthorner (approved):
Checked and confirmed it's not in Publishing E2E Tests (docker-compose.yml) 👍.

The DockerHub issue is new to me though. I'm sure you'll fix it as part of merging this. Could you explain where that check is happening? (I couldn't see it in govuk-jenkinslib.)

kevindew:
> Checked and confirmed it's not in Publishing E2E Tests (docker-compose.yml) 👍.
> 
> The DockerHub issue is new to me though. I'm sure you'll fix it as part of merging this. Could you explain where that check is happening? (I couldn't see it in govuk-jenkinslib.)

Sure, it's actually a part of govuk-app-deployment: https://github.com/alphagov/govuk-app-deployment/blob/3238b93d03c3c14e5eaa6e4fd6aad6ae8ebe9376/recipes/notify.rb#L148-L164 

If a DockerHub repo exists for the app, it'll try to tag the Docker image as `deployed-to-#{environment}` - however this will fail because the image to tag won't exist anymore.

benthorner:
Thanks, I see now 👍.

(Also forgot to say I agree with removing Wraith and the BEM docs - they really are the same as our general approach.)

PR #1058, opened 2022-03-11
Remove jQuery from highlight active section

andysellick:

## What
Removes jQuery from the only script in this application that uses it, `highlight-active-section-heading.js`.

This script highlights the menu item for the section of the page that is currently in view, on pages like https://www.gov.uk/service-manual/helping-people-to-use-your-service/making-your-service-accessible-an-introduction

There's a bit of clumsiness in this script where 3 functions do nothing but return the value of a new function, `getElementPosition`. I wanted to consolidate them all into one function, but the test depends on overriding these functions in order to pass. This isn't ideal, but any changes to improve this are outside of the scope of this PR.

## Why
We're removing jQuery from GOV.UK.

## Visual Changes
None.

### DIFF
--- a/app/assets/javascripts/modules/highlight-active-section-heading.js
+++ b/app/assets/javascripts/modules/highlight-active-section-heading.js
@@ -1,148 +1,156 @@
-;(function (Modules, root) {
-  'use strict'
+window.GOVUK = window.GOVUK || {}
+window.GOVUK.Modules = window.GOVUK.Modules || {};
 
-  var $ = root.$
-  var $window = $(root)
+(function (Modules) {
+  function HighlightActiveSectionHeading ($module) {
+    this.$module = $module
+    this._hasResized = true
+    this._hasScrolled = true
+    this._interval = 50
+    this.anchorIDs = []
+  }
 
-  Modules.HighlightActiveSectionHeading = function () {
-    var self = this
-    var _hasResized = true
-    var _hasScrolled = true
-    var _interval = 50
-    var anchorIDs = []
+  HighlightActiveSectionHeading.prototype.init = function () {
+    window.addEventListener('resize', function () { this._hasResized = true }.bind(this))
+    window.addEventListener('scroll', function () { this._hasScrolled = true }.bind(this))
 
-    self.getWindowDimensions = function () {
-      return {
-        height: $window.height(),
-        width: $window.width()
-      }
-    }
+    setInterval(this.checkResize.bind(this), this._interval)
+    setInterval(this.checkScroll.bind(this), this._interval)
 
-    self.getWindowPositions = function () {
-      return {
-        scrollTop: $window.scrollTop()
-      }
-    }
+    this.anchors = this.$module.querySelectorAll('.js-page-contents a')
+    this.getAnchors()
 
-    self.getElementOffset = function ($el) {
-      return $el.offset()
-    }
+    this.checkResize()
+    this.checkScroll()
+  }
 
-    self.start = function ($el) {
-      $window.resize(self.hasResized)
-      $window.scroll(self.hasScrolled)
-
-      setInterval(self.checkResize, _interval)
-      setInterval(self.checkScroll, _interval)
-
-      self.$anchors = $el.find('.js-page-contents a')
-      self.getAnchors()
-
-      self.checkResize()
-      self.checkScroll()
-    }
-
-    self.hasResized = function () {
-      _hasResized = true
-      return _hasResized
-    }
-
-    self.hasScrolled = function () {
-      _hasScrolled = true
-      return _hasScrolled
-    }
-
-    self.checkResize = function () {
-      if (_hasResized) {
-        _hasResized = false
-        _hasScrolled = true
-      }
-    }
-
-    self.checkScroll = function () {
-      if (_hasScrolled) {
-        _hasScrolled = false
-        var windowDimensions = self.getWindowDimensions()
-        if (windowDimensions.width <= 768) {
-          self.removeActiveItem()
-        } else {
-          self.updateActiveNavItem()
-        }
-      }
-    }
-
-    self.getAnchors = function () {
-      $.each(self.$anchors, function (i) {
-        var anchorID = $(this).attr('href')
-        // e.g. anchorIDs['#meeting-the-digital-service-standard', '#understand-your-users', '#research-continually']
-        anchorIDs.push(anchorID)
-      })
-    }
-
-    self.getHeadingPosition = function ($theID) {
-      return $theID.offset()
-    }
-
-    self.getNextHeadingPosition = function ($theNextID) {
-      return $theNextID.offset()
-    }
-
-    self.getFooterPosition = function ($theID) {
-      return $theID.offset()
-    }
-
-    self.getDistanceBetweenHeadings = function (headingPosition, nextHeadingPosition) {
-      var distanceBetweenHeadings = (nextHeadingPosition - headingPosition)
-      return distanceBetweenHeadings
-    }
-
-    self.updateActiveNavItem = function () {
-      var windowVerticalPosition = self.getWindowPositions().scrollTop
-      var footerPosition = self.getFooterPosition($('.govuk-footer'))
-
-      $.each(self.$anchors, function (i) {
-        var theID = anchorIDs[i]
-        var theNextID = anchorIDs[i + 1]
-
-        var $theID = $(theID)
-        var $theNextID = $(theNextID)
-
-        var headingPosition = self.getHeadingPosition($theID)
-
-        if (!headingPosition) {
-          return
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/modules/highlight-active-section-heading.js
```diff
@@ -1,148 +1,166 @@
...
     }
+  }
 
-    self.getHeadingPosition = function ($theID) {
-      return $theID.offset()
+  HighlightActiveSectionHeading.prototype.getAnchors = function () {
+    for (var i = 0; i < this.anchors.length; i++) {
+      var anchorID = this.anchors[i].getAttribute('href')
+      // e.g. anchorIDs['#meeting-the-digital-service-standard', '#understand-your-users', '#research-continually']
+      this.anchorIDs.push(anchorID)
```
kevindew, line 67: This has my spider senses tingling with the assumption that all href attributes are anchor links 🤔 

Can see it's in the past code though.

app/assets/javascripts/modules/highlight-active-section-heading.js
```diff
@@ -1,148 +1,166 @@
...
+  }
 
-    self.hasScrolled = function () {
-      _hasScrolled = true
-      return _hasScrolled
-    }
+  HighlightActiveSectionHeading.prototype.hasScrolled = function () {
+    this._hasScrolled = true
+    return this._hasScrolled
+  }
```
kevindew, line 35: I don't think there's any need for these functions to return - I assume the past one with jQuery was returning the boolean to not stop event handling?

To be honest I'd imagine you could relatively easily inline these with:

```
window.addEventListener('resize', function () { this._hasResized = true }.bind(this))
window.addEventListener('scroll', function () { this._hasScrolled = true }.bind(this))
```

But I understand I could be missing some testing quirk or something that'd make that hard.

app/assets/javascripts/modules/highlight-active-section-heading.js
```diff
@@ -1,148 +1,166 @@
...
+      }
+    }
+  }
 
-        if (theNextID) {
-          var nextHeadingPosition = self.getNextHeadingPosition($theNextID).top
-        }
+  HighlightActiveSectionHeading.prototype.getFooterPosition = function (element) {
+    return this.getElementPosition(element)
+  }
```
kevindew, line 116: Are we expecting this to ever get called with anything other than `document.querySelector('.govuk-footer')` ?

If not we could probably move all of 

```
    var footer = document.querySelector('.govuk-footer')
    if (footer) {
      var footerPosition = this.getFooterPosition(footer)
    }
```

into here to make the `updateActiveNavItem` method shorter.

app/assets/javascripts/modules/highlight-active-section-heading.js
```diff
@@ -1,148 +1,166 @@
...
+      var theNextID = this.anchorIDs[i + 1]
 
-    self.getDistanceBetweenHeadings = function (headingPosition, nextHeadingPosition) {
-      var distanceBetweenHeadings = (nextHeadingPosition - headingPosition)
-      return distanceBetweenHeadings
-    }
+      var $theID = document.getElementById(theID.substring(1)) // remove the # at the start
+      if (theNextID) {
+        var $theNextID = document.getElementById(theNextID.substring(1)) // remove the # at the start
+      }
```
kevindew, line 85: This code is a little confusing because it looks like the variable isn't defined unless the if is true (unless you understand all the JS hosting.

I'd suggest using a ternary to make it clearer:

```suggestion
      var $theNextID =  theNextID ? document.getElementById(theNextID) : null
```

app/assets/javascripts/modules/highlight-active-section-heading.js
```diff
@@ -1,148 +1,166 @@
...
-    }
+    for (var i = 0; i < this.anchors.length; i++) {
+      var theID = this.anchorIDs[i]
+      var theNextID = this.anchorIDs[i + 1]
 
-    self.getDistanceBetweenHeadings = function (headingPosition, nextHeadingPosition) {
-      var distanceBetweenHeadings = (nextHeadingPosition - headingPosition)
-      return distanceBetweenHeadings
-    }
+      var $theID = document.getElementById(theID.substring(1)) // remove the # at the start
```
kevindew, line 82: Do we ever need the anchor id without the "#" ? should this substring(1) be done in `getAnchors` ?

app/assets/javascripts/modules/highlight-active-section-heading.js
```diff
@@ -1,148 +1,166 @@
...
-      $.each(self.$anchors, function (i) {
-        var theID = anchorIDs[i]
-        var theNextID = anchorIDs[i + 1]
+      headingPosition = headingPosition - 53 // fix the offset from top of page
 
-        var $theID = $(theID)
-        var $theNextID = $(theNextID)
+      if (theNextID) {
+        var nextHeadingPosition = this.getNextHeadingPosition($theNextID)
+      }
```
kevindew, line 96: Similar point on tricky variable defining where a ternary might help

kevindew (commented):
Looks great - thanks Andy. I've added a few comments that are hopefully mostly scoped to just jQuery related changes.

app/assets/javascripts/modules/highlight-active-section-heading.js
andysellick: Nice optimisation, thanks for spotting, have updated.

app/assets/javascripts/modules/highlight-active-section-heading.js
andysellick: Unfortunately `getFooterPosition` is needed as it gets overridden in the test. I'll add a comment to explain, as there initially appears to be a lot of redundancy to this.

app/assets/javascripts/modules/highlight-active-section-heading.js
andysellick: Thanks, have updated.

app/assets/javascripts/modules/highlight-active-section-heading.js
andysellick: Ideally yes, but we use it with the `#` in `setActiveItem`, where we compare the value of the link href attribute `#example` with the anchor ID, which if we'd already removed the `#` would be `example`.

So either we strip it out when we collect them, then add them back in again when highlighting the links, or we keep it when we collect them, and strip it out when finding the elements. It's a bit confusing either way.

app/assets/javascripts/modules/highlight-active-section-heading.js
andysellick: Thanks, have updated.

app/assets/javascripts/modules/highlight-active-section-heading.js
kevindew: Oh sure, if we can drop it that'd be a good outcome. I do worry about the maintenance anything in here gets.

app/assets/javascripts/modules/highlight-active-section-heading.js
kevindew: I don't think moving that code into the method would prevent you overriding it?

app/assets/javascripts/modules/highlight-active-section-heading.js
kevindew: Ah fair - sorry!

app/assets/javascripts/modules/highlight-active-section-heading.js
andysellick: Hrm, for some reason this change causes the tests to fail:

```
var nextHeadingPosition = theNextID ? this.getNextHeadingPosition($theNextID) : null
```

Not sure what's going on there. I haven't got the capacity right now to dig into this - I'd like to use this optimisation, but currently can't see how.

andysellick: @kevindew can I get your thoughts on this? I'd like to optimise this but haven't got time to investigate it properly, so for the sake of a few lines I'm happy to leave it for now.

kevindew: Just been taking a look. 

The reason it fails with the change and passes previously is because, in the previous implementation if theNextID isn't a truthy value then the previous value for nextHeadingPosition will be used. Whereas with the ternary operator it'll give nextPosition a deliberate null value.

I can't really tell if the previous implementation is desired behaviour or a bug, but I think it's simpler that I approve this since it won't be changing.

kevindew (approved):

app/assets/javascripts/modules/highlight-active-section-heading.js
andysellick: Thank you!

PR #1080, opened 2022-04-19
Change layout to fix header issue

jon-kirwan:
## What

Change the `static` layout which is used for both [The Service Manual](https://www.gov.uk/service-manual) and the [Service Toolkit](https://www.gov.uk/service-toolkit) pages.

Both pages will now use a new `static` template - `gem_layout_full_width_no_footer_navigation`. Added here: https://github.com/alphagov/static/pull/2754.

All other pages use `gem_layout_no_footer_navigation`.

## Why

To fix layout issues:

- There's a blue bar (from the footer) that's showing up in the header

## Visual Changes

| Before | After |
| --- | --- |
|  |  |
|  |  |
|  |  |

## Pages to check

* /service-manual/
* /service-manual/service-standard
* /service-manual/service-toolkit

## Anything else

See https://github.com/alphagov/static/pull/2719

### DIFF
--- a/app/controllers/content_items_controller.rb
+++ b/app/controllers/content_items_controller.rb
@@ -7,7 +7,7 @@ class ContentItemsController < ApplicationController
   rescue_from GdsApi::HTTPForbidden, with: :error_403
 
   def show
-    slimmer_template :gem_layout_no_footer_navigation
+    slimmer_template layout
 
     if load_content_item
       set_expiry
@@ -21,6 +21,11 @@ class ContentItemsController < ApplicationController
   end
 
 private
+
+  def layout
+    paths = %w[service-manual service-toolkit]
+    paths.include?(params[:path]) ? "gem_layout_full_width_no_footer_navigation" : "gem_layout_no_footer_navigation"
+  end
 
   def load_content_item
     @content_item = present(


app/controllers/content_items_controller.rb
```diff
@@ -22,6 +22,11 @@ def show
 
 private
 
+  def layout
+    paths = %w[service-manual service-toolkit]
+    paths.include?(params[:path]) ? "gem_layout_full_width_no_footer_navigation" : "gem_layout_no_footer_navigation"
+  end
+
```
KludgeKML, line 29: Not sure if I'm just being dense here, but isn't this only going to work for pages where the path is exactly "service-manual" or "service-toolkit"? It looks like service-toolkit doesn't have any sub-pages, but service-manual does. Is this going to work for service-manual/design? I assume params[:path] has that whole block in it? Or am I wrong?

jon-kirwan: That's correct this will only work for pages where the path is exactly "service-manual" or "service-toolkit" (and the `gem_layout_full_width_no_footer_navigation` should only be used on the landing pages for the Service Manual - https://www.gov.uk/service-manual and https://www.gov.uk/service-toolkit).

It looks that way to me too that only the service manual has sub pages (and should use the `gem_layout_no_footer_navigation` layout).

I tested with /service-manual/helping-people-to-use-your-service and the correct template did seem to be used (screenshot included above).

KludgeKML: So in the third screenshot pair (where there's no big blue header), it's correct that they are identical? (the small blue bar appears in both)

jon-kirwan: Yep that's right. It's essentially included to show there's been no visual regressions on the sub pages because of these changes.

KludgeKML (approved):
Looks good to me. :)

